PR #4, opened 2021-03-02
Fix `business.Country` in invoice html

cpind:
Business.Country was not working:
Screenshot 2021-03-02 at 17 30 14
This fixes it. 
Tested locally: 
Screenshot 2021-03-02 at 17 31 44


paulfitz (approved):
Eek thank you!

PR #8, opened 2021-11-26
Updating widgets access levels.

berhalak:
Updating widget desired access levels.
Each widget has an additional option, `accessLevel`, which Grist uses to prompt the user for permission approval.


paulfitz (approved):
Thanks!

PR #9, opened 2021-11-29
Splitting inspect widget. Adding local dev scripts.

berhalak:
Splitting inspect widget into 2 separate widgets: Inspect Record, Inspect Table.
Allowing widget definition (inside a package.json file) to be defined as an array to produce multiple widgets from of single package.

Adding npm scripts for local testing:
- yarn run dev: main task for testing. Starts nodemon that will rebuild manifest file, and starts live-server that will automatically refresh widget on every file change.
- yarn run build: rebuilds manifest.json file.
- yarn run build:dev : rebuilds manifest.json file for local testing.
- yarn run serve: starts live server with auto-refresh.
- yarn run watch: starts nodemon that will rebuild manifest.json on every change.


index.html
```diff
@@ -7,7 +7,7 @@
   <body>
     <h2>Widgets</h2>
     <ul>
-      <li><a href="inspect/">inspect</a></li>
+      <li><a href="inspect-table/">inspect</a></li>
```
paulfitz, line 10: You can revert this now, right?

paulfitz (commented):
Thanks for the new targets - they work well for me. Just one small remaining comment.

paulfitz (approved):
Thanks!

PR #10, opened 2021-12-14
New Inspect Config widget

berhalak:
- Adding a new Inspect widget variation - "Inspect Config" that shows how one can build a configuration widget screen.
- Fixing publish a script for local testing. Replacing configurationUrl with localhost.


inspect/config.html
```diff
@@ -0,0 +1,105 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+  <meta charset="utf-8">
+  <title>Configuration</title>
+  <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
+  <style>
```
paulfitz, line 8: I'd prefer the inspect example to be as bare-bones as possible, with only essential styling. While it is useful from time to time for inspection, its main purpose is as a functional example to point people to. I don't want to see this css spreading everywhere via cut and paste, making our lives difficult down the line.

inspect/config.js
```diff
@@ -0,0 +1,28 @@
+// Handler for onOptions event, invoked after we send ready message
+grist.onOptions(options => {
+  if (options?.name) {
+    document.getElementById("textbox").value = options.name;
+  }
+});
+
+// Save button handler
+async function saveConfig() {
+  const name = document.getElementById('textbox').value;
```
dsagal, line 10: For this to serve as a good example, maybe we can factor out the reusable parts of a config right in this first PR on it. It seems to me that as long as we can designate some standard location for the config-specific CSS and JS files, we could move out the CSS from HTML to there, and move this JS file and make it generic.

To make it generic, we can save the entire form, e.g. using a function like `formDataToObj` from Grist's https://github.com/gristlabs/grist-core/blob/8853e095bb09afdfaa984de271736b6edd36114c/app/client/lib/uploads.ts#L191.

We could use this JS file for most setup, including adding the CSS to the page, and setting event handlers on everything, so that the config HTML has just `<script src="./config.js"></script>` and a plain-html form with input names that correspond to the keys of the data object, and some standard element IDs (for the form and for the buttons, if that's even needed).

dsagal (commented):
Since it's my first time following instructions in the README file, I found it unclear whether to run `yarn run build`. Eventually I realized that running it was a mistake when I intend to test locally. I think this can be clarified.

It took me some effort to get this to work. (I also had to change `grist-plugin-api.js` to be loaded from my local Grist.) But once it worked, it feels pretty nice and usable.

I agree that it's important for the recommended way to require minimal boilerplate. More on that below.

berhalak:
Closing this pull request, as the configuration screen is removed.

PR #11, opened 2021-12-31
Adding inspect config and improving local development.

berhalak:
Adding a new Inspect widget variation - "Inspect Config" that shows how to work with new options API.
Improving local testing, by rewriting grist-plugin-api.js URL.


_build/rewriteUrl.js
```diff
@@ -0,0 +1,22 @@
...
+  if (req.url && req.url.includes(".htm") && req.method === 'GET') {
+    // Overwrite Response internal _writeRaw method.
+    const _writeRaw = res._writeRaw.bind(res);
+    res._writeRaw = function (data, encoding, callback) {
+      // Make sure that the Response data is a string.
+      if (typeof data === 'string' && data) {
+        // Replace first occurrence of grist-plugin-api.js URL.
+        data = data.replace(
+          'https://docs.getgrist.com/grist-plugin-api.js',
+          'http://localhost:8080/grist-plugin-api.js    ' // extra space for length match
```
paulfitz, line 15: hmm, having used this, it is very handy - but grist-core is 8484, and I'd like this to work for everyone, and be easiest to use by grist-core users.

inspect/onConfig.html
```diff
@@ -0,0 +1,68 @@
...
+      showPanel(config ? 'main' : 'error');
+      config = config || {};
+      document.getElementById("config1").value = config.config1 || '';
+      document.getElementById("config2").value = config.config2 || '';
+      document.getElementById("setting1").innerText = config.config1 || '';
+      document.getElementById("setting2").innerText = config.config2 || '';
+      document.getElementById("app").style.display = 'block';
+    });
+    // Register configuration handler to show configuration panel.
+    grist.rpc.registerFunc('configure', () => {
```
paulfitz, line 49: I think this could be wrapped with something like `grist.onEditOptions`? Seems easier to document that way.

inspect/package.json
```diff
@@ -17,6 +17,12 @@
       "widgetId": "@gristlabs/widget-inspect#onRecords",
       "published": true,
       "accessLevel": "read table"
+    },
+    {
+      "name": "Inspect Config",
```
paulfitz, line 22: `Inspect Options` may be a better name now.

paulfitz (commented):
This feels close now!

It'd be nice to settle on consistent language for options/configuration/settings. Maybe favor options, and talk about different families of options when precision is needed? I think the grist app already has widget options, so we could talk about custom options (for the options settable within a custom widget) and interaction options (for the options related to how the widget interacts with the rest of the document, e.g. access level).

inspect/index.html
```diff
@@ -3,12 +3,12 @@
   <head>
     <meta charset="utf-8">
     <title>Grist Inspect</title>
-    <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
   </head>
   <body>
     <ul>
       <li><a href="onRecord.html">onRecord</a></li>
       <li><a href="onRecords.html">onRecords</a></li>
+      <li><a href="onConfig.html">onConfig</a></li>
```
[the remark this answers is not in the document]
berhalak: Yes.

_build/rewriteUrl.js
berhalak: I added this extra space because I'm not modifying Content-Length header (will update the comment), and this is rewriting the response just before it is sent to the client.

paulfitz: Ah, makes sense :+1:

paulfitz (approved):
Thanks!

PR #13, opened 2022-01-24
tiny page to explore public part of custom widget api

paulfitz:
A tiny helper custom widget to explore custom widget api calls. It'd be nice to grow something like this to look good and link with documentation.


inspect/api.html
```diff
@@ -5,172 +5,328 @@
...
 </head>
 
 <body>
 
+  <div style="font-weight: 400;">
+    <br />
+    <i style="color:red">*Note*</i> Do not past code that you don't trust or is yours.
+  </div>   
+  <h4>
+    Feel free to explore the API and explore the API yourself. Each snippet is fully editable. <br />
```
paulfitz, line 54: explore the API and explore the API seems redundant

inspect/api.html
```diff
@@ -5,172 +5,328 @@
...
   <h2>grist.ready()</h2>
-  <script>
-    async function gristReady(cmd) {
-      cmd.output("(already called, just needs to be called once)");
+  <div class="doc" contenteditable="true">
+    This is initial call a widget needs to make to let Grist know that
+    it is ready to process incoming events.
+  </div>
+  <form>
+    <script type="code">
```
paulfitz, line 63: Scrolling in this widget is awkward until you figure out the code boxes are capturing scroll events. Maybe they could capture scroll only if they have been clicked into?

inspect/api_deps.js
```diff
@@ -0,0 +1,821 @@
+// Typescript definitions file bundled by dts-bundle manually.
+
+const definition = `
+// Generated by dts-bundle v0.7.3
```
paulfitz, line 4: Could you write up the procedure for updating this?

berhalak (approved):
Thanks

PR #14, opened 2022-01-25
Using configuration API for labels widget

berhalak:
Print Labels widget uses configuration API to store the selected template and the number of blank cells.


paulfitz (commented):
I like this, except anyone already using print labels with the # parameter will hit a glitch where the setting suddenly changes to the default. Might be better to continue to support it, so existing documents work and existing documentation is usable?

Likewise, people with a setting saved in browser storage will see a change. I feel less worried about that, given that there's an extra dependency needed to support it, and there aren't particular guarantees, and in the end it is easy for user to fix.

paulfitz:
(I notice that actionbutton says it uses store.js, but doesn't appear to actually do so. If it did, there could have been something to clean up there too)

printlabels/printlabels.js
```diff
@@ -45,7 +42,7 @@ let app = undefined;
 let data = {
   status: 'waiting',
   labels: null,
-  template: startTemplate,
+  template: null,
```
dsagal, line 45: Is the order of calls of `onOptions` and `onRecords` guaranteed, so that template gets set by the time it's needed?

dsagal (commented):
Works beautifully! Just one concern about existing documents. The one template we have uses URL params to change the default (and that may be copied around by people who've used this template)

printlabels/printlabels.js
```diff
@@ -32,10 +32,7 @@ const templates = [{
   perPage: 80,
 }];
 
-const startTemplate =
-  findTemplate(document.location.hash.slice(1)) ||
```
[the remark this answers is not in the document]
berhalak: Yes, I will revert this change.

printlabels/printlabels.js
berhalak: You are right, there should be a template from the start, as this order is a side effect of fetching records in the onRecords handler.

printlabels/printlabels.js
```diff
@@ -121,19 +123,28 @@ function updateSize() {
 }
 
 ready(function() {
+  grist.ready({
+    requiredAccess: 'read table'
+  });
+  // Listen to configuration change.
+  grist.onOptions(options => {
+    data.template = options?.template ? findTemplate(options.template) : defaultTemplate;
```
dsagal, line 131: Shouldn't this fall back to `startTemplate`? Just checked that a doc with `#labels10` ignores that URL fragment, but changing defaultTemplate -> startTemplate here fixes that.

berhalak: Yes, sorry, last-minute refactor. I simplified this logic.

dsagal (approved):
Thanks!

PR #15, opened 2022-02-02
Using columns mapping API in exposed widgets.

berhalak:
All exposed custom widgets are not updated to use the new column API.
Change is backward compatible - each widget can still work without any mappings set.

### DIFF
--- a/actionbutton/actionbutton.js
+++ b/actionbutton/actionbutton.js
@@ -41,7 +41,7 @@ function onRecord(row, mapping) {
       throw new Error("No data row. Please add some rows");
     }
     if (!row.hasOwnProperty(data.column)) {
-      throw new Error('Need a visible column named "ActionButton". You can map custom column in the Creator Panel.');
+      throw new Error('Need a visible column named "ActionButton". You can map a custom column in the Creator Panel.');
     }
     const keys = ['button', 'description', 'actions'];
     if (!row[data.column] || keys.some(k => !row[data.column][k])) {
--- a/actionbutton/actionbutton.js
+++ b/actionbutton/actionbutton.js
@@ -32,24 +32,25 @@ async function applyActions() {
   }
 }
 
-function onRecord(row) {
+function onRecord(row, mapping) {
   try {
     data.status = '';
     data.results = null;
+    data.column = mapping ? mapping.Action : "ActionButton";
     if (!row) {
       throw new Error("No data row. Please add some rows");
     }
-    if (!row.hasOwnProperty('ActionButton')) {
-      throw new Error('Need a visible column named "ActionButton"');
+    if (!row.hasOwnProperty(data.column)) {
+      throw new Error('Need a visible column named "ActionButton". You can map custom column in the Creator Panel.');
     }
     const keys = ['button', 'description', 'actions'];
-    if (!row.ActionButton || keys.some(k => !row.ActionButton[k])) {
+    if (!row[data.column] || keys.some(k => !row[data.column][k])) {
       const allKeys = keys.map(k => JSON.stringify(k)).join(", ");
-      const missing = keys.filter(k => !row.ActionButton[k]).map(k => JSON.stringify(k)).join(", ");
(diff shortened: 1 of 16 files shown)


actionbutton/actionbutton.js
```diff
@@ -32,32 +32,33 @@ async function applyActions() {
   }
 }
 
-function onRecord(row) {
+function onRecord(row, mapping) {
```
paulfitz, line 35: Is it now possible to remove the added parameter, and simplify the `data.column = mapping ? mapping.Action : "ActionButton";` line, by using `mapColumnNames`?

clipboard/index.html
```diff
@@ -67,20 +67,32 @@
...
+      });
       grist.on('message', (e) => {
         if (e.tableId) { tableId = e.tableId; }
       });
-      grist.onRecord(function(record) {
-        delete record.id;
-        var keys = Object.keys(record);
-        if (keys.length > 1) {
-          showError("Please hide all but one of these columns from this widget: " + keys.join(", "));
+      grist.onRecord(function(record, mappings) {
```
paulfitz, line 81: Is it possible to remove a conditional from this callback by using `mapColumnNames`?

map/page.js
```diff
@@ -173,9 +177,16 @@ grist.on('message', (e) => {
 
 grist.onRecord(selectOnMap);
 if (mode !== 'single') {
-  grist.onRecords((data) => {
+  grist.onRecords((data, mappings) => {
```
paulfitz, line 180: Is it possible to simplify using `mapColumnNames`?

markdown/index.html
```diff
@@ -63,29 +63,40 @@
...
-      grist.ready();
+      grist.ready({
+        columns: [{ name: "Content", type: 'Text'}],
+        requiredAccess: 'full'
+      });
       grist.on('message', (e) => {
         if (e.tableId) { tableId = e.tableId; }
       });
-      grist.onRecord(function(record) {
+      grist.onRecord(function(record, mappings) {
```
paulfitz, line 73: Is it possible to simplify using `mapColumnNames`?

printlabels/printlabels.js
```diff
@@ -138,13 +158,27 @@ ready(function() {
     }
   })
   // Update the widget anytime the document data changes.
-  grist.onRecords(updateRecords);
+  grist.onRecords((rows, mappings) => {
+    if (mappings) {
```
paulfitz, line 162: guess what I'm going to ask :)

renderhtml/index.html
```diff
@@ -52,21 +52,26 @@
...
-      grist.ready();
+      grist.ready({
+        columns:["Html"],
+        requiredAccess: 'read table'
+      });
       grist.on('message', (e) => {
         if (e.tableId) { tableId = e.tableId; }
       });
-      grist.onRecord(function(record) {
+      grist.onRecord(function(record, mappings) {
```
paulfitz, line 62: guess what I'm going to ask :)

viewer/index.html
```diff
@@ -43,17 +43,22 @@
           el.style.display = 'block';
         }
       }
-      grist.ready();
-      grist.onRecord(function(record) {
+      grist.ready({
+        columns: [{ name: "ImageUrl", title: 'Image URL', type: 'Text'}],
+        requiredAccess: 'read table'
+      });
+      grist.onRecord(function(record, mappings) {
```
paulfitz, line 50: guess what I'm going to ask :)

markdown/index.html
berhalak: I don't think so. This widget has to update data (so it needs mappings information).
It can be done if we require explicit mapping, but it will break existing widgets - users would need to make the configuration.
What do you think? I'm in favor of breaking existing widgets, as this updates data - hence it's better to require some mapping here.

paulfitz: Got it - it is fine as it is then. Let's not break existing usage, but could you add a comment to the code saying that we're supporting an old way of mapping (by hiding all but one column) and new widgets don't need to do this - just so this code doesn't get slavishly copied by others.

paulfitz (approved):
Thanks @berhalak!

PR #16, opened 2022-03-02
Example of a widget that can read and update stock prices.

berhalak:
Example of a widget that can read and update stock prices.
Not included in the manifest yet, it should be used as an example only for now.

### DIFF
--- a/stock-monitor/index.js
+++ b/stock-monitor/index.js
@@ -1,0 +1,154 @@
+// To update Grist table, we first need to know what table we are connected to.
+// Here we will listen to all messages sent from Grist to our custom widget,
+// and if any of this message has a tableId, we will store it in global window
+// object.
+let tableId;
+grist.on("message", (data) => {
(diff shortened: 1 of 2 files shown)


stock-monitor/index.js
```diff
@@ -0,0 +1,160 @@
+// To update Grist table, we first need to know what table we are connected to.
+// Here we will listen to all messages sent from Grist to our custom widget,
+// and if any of this message has a tableId, we will store it in global window
+// object.
+let tableId;
+grist.on("message", (data) => {
```
paulfitz, line 6: (for later grist-widget api improvements - it would be nice if this information was available in a less hacky way)

paulfitz (commented):
Just copy-editing comments. Thanks for adding this!

stock-monitor/index.js
berhalak: I added this to the backlog.

PR #18, opened 2022-03-23
Adding a preview mode to the markdown editor.

berhalak:
Changing the way the save button works:
- For empty document editor is in edit mode with a save button,
- For non-empty document editor initially is in preview mode and the save button is replaced by the edit button.

Having preview mode makes links clickable, which is the main benefit of this change.

Also adding little border CSS, which makes the editor look more like a part of Grist.

### DIFF
--- a/markdown/index.html
+++ b/markdown/index.html
@@ -20,6 +20,15 @@
           padding: 20px;
           text-align: center;
       }
+      .editor-toolbar {
+        border: none
+      }
+      .EasyMDEContainer .CodeMirror {
+        border: none;
+        border-radius: 0px;
+        border-top: 1px solid #ddd;
+        border-bottom: 1px solid #ddd
+      }
     </style>
   </head>
   <body>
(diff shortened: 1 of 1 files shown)


paulfitz:
Tried it out, it works nicely and seems better in most respects. Do you know if there's a keyboard shortcut for editing? It is a little tedious to click on the little button if switching frequently from row to row.

paulfitz:
(I made a tiny change to your pull request in order to trigger a preview deployment I set up - handy to have a link to your widget I can use without needing to download and serve locally).

berhalak:
> Tried it out, it works nicely and seems better in most respects. Do you know if there's a keyboard shortcut for editing? It is a little tedious to click on the little button if switching frequently from row to row.

Added 3 keyboard shortcuts
- Enter or space to go to the edit mode
- Escape to cancel
- Ctrl/Cmd + S to save

Can you test Cmd + S on mac?

paulfitz (approved):

PR #20, opened 2022-04-01
Map: keep popup open on zoom

paulfitz:
When the map is showing many locations, the selected location can be lost through clustering as the user zooms in. This adds a hacky way to keep that location open.

### DIFF
--- a/map/page.js
+++ b/map/page.js
@@ -20,6 +20,8 @@ const Geocode = 'Geocode';
 const Address = 'Address';
 // Optional - but 
 const GeocodedAddress = 'GeocodedAddress';
+let lastRecord;
+let lastRecords;
 
 const geocoder = L.Control.Geocoder && L.Control.Geocoder.nominatim();
 if (URLSearchParams && location.search && geocoder) {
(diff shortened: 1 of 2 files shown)


map/page.js
```diff
@@ -176,13 +178,23 @@ function updateMap(data) {
...
+      if (!marker._icon) { marker.__parent.spiderfy(); }
+      marker.openPopup();
+    }
   }
+  map.on('zoomend', () => {
+    // Should reshow marker if it has been lost, but I didn't find a good
+    // event to trigger that exactly. A small timeout seems to work :-(
+    // TODO: find a better way; also, if user has changed selection within
+    // the map we should respect that.
+    setTimeout(makeSureSelectedMarkerIsShown, 500);
```
berhalak, line 194: This makes the map move always to show the marker, I found that this fixes that issue.

`171: marker.bindPopup(title, {autoPan: false});`

paulfitz: Hmm, Natalie seemed to specifically like that "feature" - `I like that if you zoom in and it's no longer in the window, it adjusts your view of the map to show you the selected location` - do you see a problem with it?

berhalak: Ah ok, I thought that it would be seen as a bug - because you can't zoom anywhere, but maybe it is ok, this map's purpose is to show those points after all.

berhalak (approved):

PR #21, opened 2022-04-12
New widget - Notepad - rich text editor.

berhalak:
New widget: "Notepad". 'WYSIWYG'-like editor similar to Markdown with autosave and images.

### DIFF
--- a/notepad/index.css
+++ b/notepad/index.css
@@ -1,0 +1,26 @@
+.ql-toolbar.ql-snow {
+  border-top: none;
+  border-left: none;
+  border-right: none;
+}
+
+.ql-container.ql-snow {
+  border: none;
+  height: auto;
+  flex: 1;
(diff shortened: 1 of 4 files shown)


berhalak:
I discussed this with Natalie and we decided that it would be useful to add it for Thursday's webinar.

paulfitz:
Cute! Nice to have a way to include images. Encoding them in the cell may give @dsagal qualms about data engine performance, but it seems fine for small-scale use-cases. I played around with the widget, and moving from row to row felt a bit hit and miss, but I think it was because I had some material in the column to be edited that was not in "{ops: ..." serialization format. Would it be possible to "upgrade" such material?

anaisconce:
Looks great! I'm having fun with it and eager to have it!

paulfitz (approved):
Thanks @berhalak !

PR #24, opened 2022-04-23
Add Dropbox Embedder custom widget 

dsagal:
Use can place [shared links](https://help.dropbox.com/files-folders/share/view-only-access) to Dropbox files or folders into a cell, and the widget will show a preview. For example:
Screen Shot 2022-04-23 at 12 20 09 AM


berhalak (commented):
Works great, thanks. Just a suggestion to parse Grist links.

berhalak (approved):
Thanks.

PR #28, opened 2022-05-06
Simple widget for embedding video files

berhalak:
Simple widget for embedding video files.
Supports videos from popular platforms and direct URLs to media files.

### DIFF
--- a/rendervideo/index.html
+++ b/rendervideo/index.html
@@ -1,0 +1,134 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="utf-8">
+  <title>Grist Player</title>
+  <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
+  <style>
+    html,
+    body {
+      width: 100%;
+      height: 100vh;
+      padding: 0;
+      margin: 0;
+    }
+
+    #error {
+      display: none;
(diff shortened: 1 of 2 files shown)


rendervideo/index.html
```diff
@@ -0,0 +1,134 @@
...
+    function singleColumn(record) {
+      const columns = Object.keys(record || {}).filter(k => k !== 'id');
+      return columns.length === 1 ? record[columns[0]] : undefined;
+    }
+    grist.onRecord(function(record) {
+      // If user picked all columns, this helper function will return a mapped record.
+      const mapped = grist.mapColumnNames(record);
+      // We will fallback to reading a value from a single column to
+      // support old way of mapping (exposing only a single column).
+      // New widgets should only check if mapped object is truthy.
```
paulfitz, line 119: I don't think you need this fallback.

paulfitz (commented):
Tried it out, it works well. Contains tweaks specifically for youtube, which is reasonable enough given dominance. Seems like it might be a bit frustrating if custom section has plenty of space to show a portrait video, but it gets squeezed down because of aspect ratio assumption? Not a big deal for a V1.

paulfitz (approved):
Approving this since it is pretty good and we can iterate.

PR #31, opened 2022-05-18
play with support for a print page title

paulfitz:
This is just an experiment for a question from a user.


paulfitz:
Demo using this at https://docs.getgrist.com/uqqdyuiQCHEi/Invoicing-copy/m/fork - added a Title column and made it visible to the widget.

PR #34, opened 2022-08-05
[Notepad] Adding font size options and image resize plugin

berhalak:
Adding to new options for notepad widget:
1. Image resize plugin (allows resizing pasted images)
2. Revealing font size control


notepad/index.js
```diff
@@ -6,6 +6,7 @@ const toolbarOptions = [
   ['blockquote', 'code-block'],
 
   [{ 'header': [1, 2, 3, 4, 5, 6, false] }],
+  [{ 'size': ['small', false, 'large', 'huge'] }],  // custom dropdown
```
paulfitz, line 9: The default text in the header and size dropdowns appears to be the same ("normal") meaning unless you memorize order you don't know what you'll be controlling until you click. Any way to do better? No big deal if there's nothing easy to do here.

notepad/image-resize.min.js
```diff
@@ -0,0 +1,15 @@
+// Credits to https://github.com/Etoile984816138/quill-image-resize-module
```
paulfitz, line 1: Can you also mention the license under which we are using it (MIT)

paulfitz (approved):
Looks good! Two small comments.

paulfitz (approved):

PR #35, opened 2022-09-25
Keeping focus within Grist document when switching to an empty editor.

berhalak:
Markdown widget was stealing focus from Grist when the user switched between filled and empty rows.
This change is addressing this issue https://community.getgrist.com/t/issue-markdown-widget-stealing-focus/1450.


paulfitz (approved):
Thanks @berhalak !

PR #36, opened 2022-10-27
Include ts-interface-checker in dts bundle + export raw dts file

arnaudpeich:
I heard about this definition files bundler [here](https://github.com/gristlabs/grist-core/pull/328#pullrequestreview-1156653425). While running it I noticed it was missing another external library, `ts-interface-checker`, which was probably added after this script was written. This PR adds it.

In the second commit, I also write out a raw definition file `grist-plugin-api.d.ts` which I use to get IntelliSense while writing widgets.

The way I use it is that I copy the definition file to my widget repo, remove the quotes around `'grist'` in `declare module 'grist'` to make it internal, and reference it with:
```ts
/// <reference path="path/to/grist-plugin-api.d.ts" />
```

There's probably a smarter way to do this, I'm still poking around.
@berhalak I am curious to know what was the original use for this script 🙂
Especially why the definition is extracted into a string in a JS file.


berhalak (approved):
Thanks @arnaudpeich for this update.
This file is used in the `Inspect Api` widget, you can see it on hosted grist or by using [this link](https://gristlabs.github.io/grist-widget/inspect/api.html) as a `Custom URL` for custom widget. We are using there monaco editor and this bundled types are imported on the client side to provide some IntelliSense.
Those types are also used in this [fiddle widget](https://berhalak.github.io/my-widgets/fiddle/index.html), but it is just an experiment for now and it needs some polish :).

inspect/api_deps.js
```diff
@@ -24,22 +26,148 @@ declare module 'grist' {
...
+        * document. For example, in a custom widget for a table with a 'Photos' column
+        * containing attachments, the following code will update the 'src' of an
+        * image with id 'the_image' to show the attachment:
+        * '''js
+        * grist.onRecord(async (record) => {
+        *   const tokenInfo = await grist.docApi.getAccessToken({readOnly: true});
+        *   const img = document.getElementById('the_image');
+        *   const id = record.Photos[0];  // get an id of an attachment - there could be several
+        *                                 // in a cell, for this example we just take the first.
+        *   const src = '${tokenInfo.baseUrl}/attachments/${id}/download?auth=${tokenInfo.token}';
```
berhalak, line 114: I just noticed it. @arnaudpeich can you add a simple replace to the generate.js file (line 56) to escape these characters in the string? Something like this would be enough:

`${contents.replace(/\$/g, "\\$")}`

berhalak:
Thank you!.

PR #37, opened 2022-11-04
Custom plotly chart widget

alexmojaki:
Provides a simple code editor where you can write the body of a JS function that accepts `table` (an object of columns) and returns `[data, layout]` which are passed to plotly.

In this particular case it might make more sense to use regular formulas to generate `[data, layout]` as JSON instead of writing JS code. But the idea was to generalise this widget to other plotting libraries, including Python libraries running with Pyodide. Didn't get around to that.


berhalak:
Closing for now, can be reopened later.

PR #38, opened 2022-11-24
PivotTable

jperon:
This widget builds a [pivot table](https://pivottable.js.org/examples/) from the source data.

Just setting its url gives a pivot table without any row / column definition.

The settings may be adjusted by passing url parameters, for example (with data similar to [this example](https://pivottable.js.org/examples/mps_agg.html)):

```url
https://example.com/path/to/pivottable?rows=Party,Province&cols=Gender&aggregatorName=Average&rendererName=Col%20Heatmap
```

### DIFF
--- a/pivottable/index.js
+++ b/pivottable/index.js
@@ -1,0 +1,57 @@
+window.onerror = (err) => {
+  console.trace();
+  alert(String(err));
+};
+
+grist.ready({
+  requiredAccess: 'read table'
+});
+
+const urlParams = new URLSearchParams(window.location.search);
+const rows = urlParams.get('rows')?.split(',');
+const cols = urlParams.get('cols')?.split(',');
+const vals = urlParams.get('vals')?.split(',');
+const aggregatorName = urlParams.get('aggregatorName');
+const rendererName = urlParams.get('rendererName');
+
+function wavg (n) {
+  if (!n) { return; }
+  n = n.filter(([note]) => typeof (note) === 'number');
+  if (n.length) { return n.map(([note, coef]) => note * coef).reduce((a, b) => a + b) / n.map(([_note, coef]) => coef).reduce((a, b) => a + b); }
+}
+
+function weightedAverage ([val, coef]) {
+  return (_data, _rowKey, _colKey) => ({
+    values: [],
+    push: function (rec) { this.values.push([rec[val], rec[coef]]); },
+    value: function () { return wavg(this.values); },
+    format: function (x) { return (Math.round(x * 100) / 100).toFixed(2); },
+    numInputs: 2
+  });
+}
+
+const aggregators = {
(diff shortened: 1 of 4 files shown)


jperon:
[Here is](https://docs.getgrist.com/wJgMXWTKnu5u/Notes) a sample document.

dsagal:
This is so cool!

jperon:
Really, what is cool is the work both you and PivotTable have done: I’d never believed it would be so easy to build such an interface with 50 lines of js.

pivottable/index.js
```diff
@@ -0,0 +1,57 @@
...
+    numInputs: 2
+  });
+}
+
+const aggregators = {
+  'Weighted Average': weightedAverage
+};
+
+grist.onRecords(rec => {
+  console.log(rec);
```
berhalak, line 38: You can use Grist API to save and restore this settings. For example:

```js
grist.onRecords(async rec => {
  const { rows, cols, vals, aggregatorName, rendererName } = await grist.getOption('settings') ?? {};
  let initialRender = true;
  $('#table').pivotUI(
    rec,
    {
      rows,
      cols,
      vals,
      onRefresh: function (config) {
        if (initialRender) {
          initialRender = false;
          return;
        }
        const { rows, cols, vals, aggregatorName, rendererName } = config;
        grist.setOption('settings', { rows, cols, vals, aggregatorName, rendererName });
      },
```

pivottable/index.js
```diff
@@ -0,0 +1,57 @@
...
+  $('#table').pivotUI(
+    rec,
+    {
+      rows,
+      cols,
+      vals,
+      aggregatorName,
+      rendererName,
+      aggregators: $.extend(aggregators, $.pivotUtilities.aggregators),
```
berhalak, line 47: What do you think about reversing the order, as built-in aggregations seem to be more commonly used by others?

```js
aggregators: $.extend($.pivotUtilities.aggregators, aggregators),
```

berhalak (changes requested):
Thank you for your contribution. This is truly a very cool widget to have in Grist.

pivottable/index.js
jperon: Done. Should I keep the url parameters as an alternative, or delete them ?

berhalak (approved):
Thanks, looks good. Regarding the dark theme, custom widgets currently don't have access to this setting, but we should fix it in the near future. For now, you can try to use custom [CSS styles](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

```css
@media (prefers-color-scheme: dark) {
  body {
    background-color: #32323f;
  }
}
```

jperon:
Just pushed the change. The table itself stays with its colors, but at least the options get the correct theme.

PR #39, opened 2023-01-09
Publish PivotTable

jperon:
I hadn’t seen that file in other widgets.


paulfitz:
Thanks for doing this @jperon !

PR #41, opened 2023-01-19
Improvements to the markdown widget

dsagal:
- JS code moved from index.html into a separate JS file
- Styling updated to look cleaner (more like Github's markdown styling) rather than default styles.
- Toolbar only shown in full in edit mode
- In read mode only the "edit" button is overlaid on the content, for clean rendering; and hidden when printing.
- Key behavior unchanged (Enter or Space to start editing, Cmd-S or Escape to stop)

Some previews:

Before
Screenshot 2023-01-19 at 9 30 09 AM

After
Screenshot 2023-01-19 at 9 29 29 AM

### DIFF
--- a/markdown/index.html
+++ b/markdown/index.html
@@ -23,6 +23,22 @@
         border-radius: 4px;
         color: black;
       }
+      blockquote {
+        border-left: 4px solid #ddd;
+        margin: 0;
+        padding: 0 1rem;
+        color: #666;
+      }
+      .editor-preview pre {
+        background-color: #eef3ff;
+        padding: 8px;
+        border-radius: 4px;
+      }
+      .editor-preview pre code {
+        background-color: unset;
+        padding: unset;
+        border-radius: unset;
+      }
       hr {
         margin-top: 2rem;
         margin-bottom: 2rem;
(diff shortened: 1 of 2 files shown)


markdown/index.html
```diff
@@ -6,190 +6,121 @@
...
-            editMode();
-            ev.preventDefault();
-            return;
-          }
+      .toolbar-read-mode {
+        position: absolute;
+        top: 4px;
+        right: 4px;
+        z-index: 100;
+        background-color: white;
```
berhalak, line 90:
```suggestion
        background-color: #ffffff80;
```
What do you think about making it transparent a little bit, to show a text behind it when the first line is long enough?

berhalak (approved):
Thanks @dsagal, looks great.

markdown/index.html
dsagal: Good idea. I am thinking also of giving it a green border, so it's a bit more noticeable, wdyt?

Screenshot 2023-01-19 at 12 23 45 PM

berhalak: Yes, it looks great.

berhalak (approved):

PR #45, opened 2023-05-18
Avoid negative zeroes in invoices, and show $0.00 when that is a legitimate total

dsagal:


invoices/invoice.js
```diff
@@ -70,7 +70,12 @@ let app = undefined;
 
 Vue.filter('currency', formatNumberAsUSD)
 function formatNumberAsUSD(value) {
-  if (!value) { return '—'; }
+  if (typeof value !== "number") {
+    return value || '—';      // null and undefined would be shown as a dash.
```
berhalak, line 74:
```suggestion
    return value || '—';      // falsy value would be shown as a dash.
```

Or you can use the `??` operator that will target only null/undefined.

berhalak (approved):
Thanks. Two small suggestions.

PR #46, opened 2023-05-24
show progress from "1 to N" instead of "0 to N"

wunter8:
By showing the list index to the user, for a quiz with 10 questions, the progress bar starts at "0 of 10" and the last card shows "9 of 10", so it seems like you never get to the final card.

This changes it to start at "1 of 10" and end on "10 of 10".


dsagal (approved):
Thank you!

PR #47, opened 2023-05-24
show progress bar as full when on last card

wunter8:
The previous PR fixed the text to go from 1-N instead of 0-N.

This PR fixes the progress bar so that when you're on the final card, it is 100% full.

Sorry I didn't catch it in the previous PR!

### DIFF
--- a/flashcards/index.js
+++ b/flashcards/index.js
@@ -42,7 +42,7 @@ function goNext(step) {
     if (at < 0) { at = 0; }
   }
   at = at % questions.length;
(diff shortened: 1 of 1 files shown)


dsagal:
I was wondering about that for the first PR, but it seemed actually a reasonable distinction to keep: the bar shows how much is completed (so starts with 0%), and number shows the number of the current card. I don't feel strongly either way, but just pointing out that it doesn't feel broken. If you still want to change, let me know, I'm happy to approve.

wunter8:
I think this looks funny:

dsagal (approved):
The new version works beautifully, thanks!

PR #49, opened 2023-07-13
Updating css for printing in notepad widget

berhalak:
The notepad widget with the Quill editor didn't support printing styles. This adds some basic support that should work in most cases.


georgegevoian (approved):
Thanks @berhalak!

PR #50, opened 2023-07-14
Enable theme configuration for Notepad widget

samkhal:
The default Quill theme's toolbar takes up a lot of space, so I wanted to allow selecting a different theme. The Bubble theme only shows the toolbar as a tooltip when text is selected. This PR adds a config screen where the theme can be changed.

As a side-note: this method of setting widget options seems rather clunky. Has there been any discussion about having widgets expose options directly in the sidebar?

### DIFF
--- a/notepad/index.css
+++ b/notepad/index.css
@@ -17,13 +17,15 @@ body {
   font-size: 13px;
 }
 
-body {
+.editor-container{
+  height: 100%;
   display: flex;
   flex-direction: column;
   font-family: -apple-system, BlinkMacSystemFont, Segoe UI, Helvetica, Arial, sans-serif,
     Apple Color Emoji, Segoe UI Emoji, Segoe UI Symbol;
   overflow: hidden;
 }
+
 .ql-snow .ql-picker.ql-size .ql-picker-label::before {
   content: 'Regular'
 }
--- a/notepad/index.html
+++ b/notepad/index.html
@@ -14,7 +14,21 @@
 </head>
 
 <body>
-  <div id="editor">
+  <form id="configuration">
+    <h2>Notepad Widget Configuration</h1>
+    <div>
+      <label>Select Quill theme:
+        <select name="quillTheme" id="quillTheme">
+          <option value="snow">Snow (format via toolbar)</option>
+          <option value="bubble">Bubble (format via tooltip)</option>
+        </select>
+      </label>
+    </div>
+    <div>
+      <input type="submit" value="Save">
+    </div>
+  </form>
+  <div id="editor" class="editor-container">
   </div>
   <script src="index.js"></script>
 </body>
--- a/notepad/index.js
+++ b/notepad/index.js
@@ -1,6 +1,8 @@
 const {Subject} = rxjs;
 const {debounceTime} = rxjs.operators;
 
+const defaultTheme = "snow";
+
 const toolbarOptions = [
   ['bold', 'italic', 'underline', 'strike'],        // toggled buttons
   ['blockquote', 'code-block'],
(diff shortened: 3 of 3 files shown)


JakubSerafin:
configuration page seems to be very basic and kind of ugly, but it do the job.

paulfitz:
Thanks @samkhal ! There's definitely been discussion of a more integrated way to configure custom widgets. We've taken some steps in the past, like a way to map columns. Last time we looked at this we got a bit bogged down in designs that all introduced some new kind of clunkiness. We are starting to think again about integrating custom widgets further with the main app. If you have a vision of how it could work (and time and energy 🙂) mockups are a good way to get a conversation going.

samkhal:
Coming from Airtable, it's so nice to be able to contribute so easily! Love this project.

@JakubSerafin it's definitely ugly. I'm no web designer, hence why I'd like there to be a nice default interface for widget config.

@paulfitz Happy to make an MVP mockup, I'll open a new issue. Are previous discussions available somewhere so I don't retread old ground?

paulfitz:
@samkhal here's some design work, part of which we did (in some way), a lot of which we did not do:

Per-section widget configuration

Custom widgets will soon be easier for user to pick - but then actually setting them up to do what user wants is still complicated. Currently configuration, such as it is, is done by adding query parameters to a url, and/or carefully selecting which columns are shared with widget, and/or a convention in column naming.

Goals:

* To provide a user-friendly way to provide configuration that is particular to a custom widget instance.
* Ideally, making custom widgets closer to native widgets in user experience (particularly for graphing).
* Ideally, keeping custom widget development affordable (at least some useful widgets can be made with a few dozen lines of html+css+js, within reach of web devs, or non-developers stretching themselves)

Proposal 1: Custom configuration on the Creator Panel

We can expose custom widget configuration on the creator panel in the right tab called "Settings" . The new tab will provide a standard place to show the widget configuration and offer a common look and feel familiar to Grist clients.

The configuration tab can be created in two ways:

1. [iframe] As an embedded iframe that shows configuration page prepared by the widget developer (external webpage)
2. [form builder] As custom panel that Grist will render based on the configuration or api calls exposed by Grist and provided by widget developer.

custom-config1

Examples of widgets that would benefit from creator panel configuration

1. Charts, as this is the obvious thing we should move outside from Grist.
2. All widgets that would require basic access settings, like API Keys, external URLs to use, branding information, that should be hidden from end users:
    1. Integrations with some eCommerce platform, to show product details, create packages, print invoices.
    2. Skins and branding, showing custom logo, custom name, custom help pages.
4. Custom external permissions, that should be defined only by a document owner.

Option 1: [iframe] Custom configuration as second widget on iframe

Extra iframe available as a new tab gives widget developer a lot of freedom, to structure and implement the configuration options that is best suit for particular widget.

* Configuration is set in an iframe embedded in creator panel.
* Widget-iframe and configuration-iframe coordinate via json passed from one to the other by Grist.
* Configuration is stored inside view section metadata (_grist_Views_section.options)
* Configuration-iframe receives information about records/columns, like the widget-iframe does (might be needed for populating options). Essentially making it custom widget itself.
* Widget manifest defines URL of a page to render inside the iframe.

Advantages:

* More control for developers, they control layout, structure and behavior, but it require more skills.

Disadvantages:

* We don’t control layout, structure or behavior of configuration editor
    * To provide common UI/UX experience we can provide an npm package with custom components (native webcomponents or ReactJS/VueJS library) that could be use to build the configuration. It will require little effort from the widget developer, to make the configuration look like a part of Grist.
        Cons: This can be problematic if we decide to change the user interface, as it will require an upgrade or a change to the widget source code. We can be addressed by carefully designing those web components that are UI agnostics and not embedding them in the custom widget source code, but rather importing those libraries from grist-static.
* More complex to implement for non developers.

Option 2: [form builder] Custom configuration via forms builder

Once a particular view/section of a page has been set as a particular custom widget, it may have some extra options the user can set. It makes sense for those options to be set in the creator panel, like other widgets. It also makes sense for the widget to control those options. Ideally, layout, styling, and behavior of the creator panel would remain tightly controlled, since it is a relatively small space.

Perhaps the widget could have access to the following new methods and callbacks:

* onOptionChange - called with json payload whenever relevant configuration in creator panel changes
* getOptions - can be called at any time to get what onOptionChange would provide
* a set of methods that control (with hands off) a “configuration canvas”
    * addChoice(id, {...choiceOptions...})
        * Adds a new choice option to configuration. Identified with id. Will be populated with any value persisted for that id for that section by Grist. Choice options may be record-and-column-data-dependent, since the widget will have received information about records and columns.
    * .... other add methods, hopefully not too many ...
    * reset()
        * Clears canvas (probably so it can be rejiggered, with new calls to addChoice etc - maybe some choice opened up a new branch of options). Doesn’t clear configuration.

Configuration is stored inside view section metadata (_grist_Views_section.options)

Advantages

* We retain control over layout, styling, behavior of configuration editor,
* Widget developer doesn’t need to worry about layout, styling, behavior of configuration editor,
* Lower barrier to entry,
* Pretty flexible and dynamic

Disadvantages

* We’re stuck dealing with offering all the kinds of config a widget might need
    * This could get out of hand
    * Mitigation: for crazy stuff, the widget maker could always do the UI within the widget (we’d just need to make sure they can add on that config state without a visible element in side panel - like hidden inputs in html forms)
* Rewriting the config options could be a little clunky as number of options get large
* There’ll be a constant stream of other feature requests, e.g. grouping config options, tabs, etc

Feels like a fairly simple implementation would enable a lot of widgets, but that it would eventually be pushed to be a form builder project - which is on our roadmap, maybe there could be synergy.


Proposal 2: Custom configuration on the popup

Instead of using the creator panel, we can expose widget configuration in a modal popup (similar to Import) that will host the iframe with a custom configuration page. When the custom widget defines a configuration section, we will show an additional button in the creator panel.

Integration with Grist is the same as with Proposal 1, the difference is only in the UI.

custom-config2


Advantages

* There is more available space for custom configuration. Widgets like Custom Imports, Embedded external apps, Print templates will require a larger space.
* UI/UX doesn’t need to be consistent with other parts of Grist,
* Creator panel remains unchanged and is used only for Grist configuration.

Disadvantages

* For smaller widgets with few options, a big popup might not look right. Custom Widget manifest might need to specify some desired popup dimensions

PR #51, opened 2023-07-17
get set up for systematically testing custom widgets

paulfitz:
This is a start at a test harness for the grist-widget repo. The `test/gristWebDriverUtils.ts` file depends on a small refactor of `grist-core` (see https://github.com/gristlabs/grist-core/pull/576). To exercise the harness, I wrote a small test for `inspect` and `flashcards`.

We have an embarrassment of riches in ways to have a Grist page to test with (grist-core, grist-electron, grist-static). I went with Grist in docker because it is most mature but am not committed to it.

The harness is a bit rough and ready in places - just making a start.


georgegevoian (approved):
Looks good to me. Thanks @paulfitz.

PR #52, opened 2023-07-21
make manifest.json available for use from browsers

paulfitz:
It would be useful to have a copy of `manifest.json` that is usable from browsers (for grist-static). GitHub doesn't set a CORS origin wildcard for releases, but it does for GitHub Pages. Also jsdelivr can mirror branches (but not releases). So this adds a build step to make a branch that contains `master` plus the generated `manifest.json` file.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,7 +9,9 @@ jobs:
   release:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


paulfitz:
(build step tested on https://github.com/paulfitz/grist-widget)

JakubSerafin (approved):
thanks @paulfitz and Paul's Grist Bot :)

PR #53, opened 2023-07-22
Map Widget Improvements

jvorob:
Changed tile source so doesn't break at max zoom
Map pin for selected row is highlighted green
Clustering behavior improved
No longer jumps around on zoom
Selected a row zooms to it, instead of spiderfying 
Minor other fixes

Demo here:
[Old Map Widget](https://docs.getgrist.com/7ukHUbCHxTgH/Wedding-Planner-Demo-Vanilla/p/17)
[Improved Map Widget](https://docs.getgrist.com/7kguGEPS6quW/Wedding-Planner-Demo-Maptweaks/p/17)


map/page.js
```diff
@@ -18,11 +18,76 @@ const Latitude = "Latitude";
 const Geocode = 'Geocode';
 // Optional - but required for geocoding. Field with address to find (might be formula)
 const Address = 'Address';
-// Optional - but 
+// Optional - but
```
paulfitz, line 21: Do you have a guess for what this comment was trying to say? but ... ditto?

map/page.js
```diff
@@ -138,10 +203,19 @@ function updateMap(data) {
...
     return;
   }
-  const tiles = L.tileLayer('//server.arcgisonline.com/ArcGIS/rest/services/NatGeo_World_Map/MapServer/tile/{z}/{y}/{x}', {
-    maxZoom: 18,
-    attribution: 'Tiles &copy; Esri &mdash; National Geographic, Esri, DeLorme, NAVTEQ, UNEP-WCMC, USGS, NASA, ESA, METI, NRCAN, GEBCO, NOAA, iPC'
+
+
+// Map tile source:
+//    https://leaflet-extras.github.io/leaflet-providers/preview/
+//    Old source was natgeo world map, but that only has data up to zoom 16
```
paulfitz, line 210: Looks better. If it were easier to add options, I'd add an option here, since existing map users might be used to the existing map style. Hmm, I think we can go ahead and make the change, given that grist-widget is firmly at pre-release version numbers. I tagged the current state of the repo and if anyone doesn't like our change I think https://cdn.jsdelivr.net/gh/gristlabs/grist-widget@v0.0.2/map/index.html would work for them and give us time to add configuration.

map/page.js
```diff
@@ -165,35 +265,47 @@ function updateMap(data) {
...
     }
   }
-  map.on('zoomend', () => {
-    // Should reshow marker if it has been lost, but I didn't find a good
-    // event to trigger that exactly. A small timeout seems to work :-(
-    // TODO: find a better way; also, if user has changed selection within
-    // the map we should respect that.
-    setTimeout(makeSureSelectedMarkerIsShown, 500);
-  });
+  // === Removing this: jumping around every time you zoom makes it really hard to use
```
paulfitz, line 297: I'm ok with deleting this.

paulfitz (commented):
Feels like a significant improvement to me. Old behavior was quite janky. @anaisconce I think you said you wanted to take a look also?

map/page.js
paulfitz: (Update, no, jsdelivr doesn't work, html is served as plain text. Would need to push v0.0.2 to grist-static.com or someplace.)

map/page.js
jvorob: Something like "Optional - but useful for geocoding. Blank field which map uses to store last geocoded Address. Enables map widget to automatically update the geocoding if Address is changed"

These comments look systematic, do these get parsed and sent to grist UI somewhere? If so my description might be too long.

paulfitz (approved):
Thanks @jvorob !

PR #54, opened 2023-07-25
add custom widget supporting memes template

paulfitz:
This adds in a custom widget that was lying around in my personal fork, and got used to support a Grist template:

  https://templates.getgrist.com/gtzQwTXkgzFG/Meme-Generator

It isn't great code but it works, so just adding it unchanged other than some comments and formatting. I occasionally zap my personal fork so it would be safer not to rely on it.

### DIFF
--- a/memes/index.html
+++ b/memes/index.html
@@ -1,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Grist Memes</title>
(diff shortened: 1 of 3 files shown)


paulfitz:
Confirmed that forking the template and replacing the custom widget URL works.

JakubSerafin (approved):
because "It isn't great code but it works, so just adding it unchanged other than some comments and formatting" i'm not focusing on code quality here. Code is understandable and it's enough :) thanks!

PR #59, opened 2023-09-05
Feature/calendar month name

JakubSerafin:
Calendar view was missing name of the month and the year that user is actually browsing, so I've added it. .

Test for that functionality added.

Bit of stilling (baseline, arrows) improved in UI.

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -26,8 +26,8 @@ function isRecordValid(record) {
   typeof record.title === 'string'
 }
 
-function getMonthName(){
-  return calendarHandler.calendar.getDate().d.d.toLocaleString('en-us', { month: 'long', year:'numeric' })
+function getMonthName() {
+  return calendarHandler.calendar.getDate().toDate().toLocaleString('en-us', { month: 'long', year: 'numeric' })
 }
 
 class CalendarHandler {
@@ -89,8 +89,11 @@ class CalendarHandler {
       this.calendar.updateEvent(record.id, CALENDAR_NAME, {backgroundColor: CalendarHandler._selectedColor});
       this._selectedRecordId = record.id;
       this.calendar.setDate(record.startDate);
+      updateUIAfterNavigation();
       if (this.calendar.getViewName() !== 'month') {
-        //Scoll to the middle of the event, if it's not month view
+        //Scroll to the middle of the event if it's not month view.
+        //In some cases, an event is not visible even if a valid day is focused - for example, when event is in the
+        //last hour of the day, so to make it visible, we need to scroll to the middle of the event.
         const dom = document.querySelector('.toastui-calendar-time');
         const middleHour = record.startDate.getHours()
           + (record.endDate.getHours() - record.startDate.getHours()) / 2;
@@ -108,19 +111,19 @@ class CalendarHandler {
   // navigate to the previous time period
   calendarPrevious() {
     this.calendar.prev();
-    setMonthNameInDom();
+    updateUIAfterNavigation();
   }
 
   // navigate to the next time period
   calendarNext() {
     this.calendar.next();
-    setMonthNameInDom();
+    updateUIAfterNavigation();
   }
 
   //navigate to today
   calendarToday() {
     this.calendar.today();
-    setMonthNameInDom();
+    updateUIAfterNavigation();
   }
 
   // update calendar events based on the collection of records from the grist table.
@@ -195,7 +198,8 @@ function getGristOptions() {
 }
 
 
-function setMonthNameInDom(){
+function updateUIAfterNavigation(){
+  //update name of the month and year displaed on the top of the widget
   document.getElementById('calendar-title').innerText = getMonthName();
 }
 // let's subscribe to all the events that we need
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -88,10 +88,13 @@ class CalendarHandler {
       this.calendar.updateEvent(record.id, CALENDAR_NAME, {backgroundColor: CalendarHandler._selectedColor});
       this._selectedRecordId = record.id;
       this.calendar.setDate(record.startDate);
-      const dom = document.querySelector('.toastui-calendar-time');
-      const middleHour = record.startDate.getHours()
-        + (record.endDate.getHours() - record.startDate.getHours()) / 2;
-      dom.scrollTo({top: (dom.clientHeight / 24) * middleHour, behavior: 'smooth'});
+      if (this.calendar.getViewName() !== 'month') {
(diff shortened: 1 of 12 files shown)


calendar/page.js
```diff
@@ -26,12 +26,17 @@ function isRecordValid(record) {
   typeof record.title === 'string'
 }
 
+function getMonthName(){
+  return calendarHandler.calendar.getDate().d.d.toLocaleString('en-us', { month: 'long', year:'numeric' })
```
paulfitz, line 30: what is .d.d?

paulfitz: most grist code leaves a space after a colon

paulfitz: and a space before a curly brace starting a function (like `getMonthName() {`)

calendar/page.js
```diff
@@ -82,31 +88,38 @@ class CalendarHandler {
       this.calendar.updateEvent(record.id, CALENDAR_NAME, {backgroundColor: CalendarHandler._selectedColor});
       this._selectedRecordId = record.id;
       this.calendar.setDate(record.startDate);
-      const dom = document.querySelector('.toastui-calendar-time');
-      const middleHour = record.startDate.getHours()
-        + (record.endDate.getHours() - record.startDate.getHours()) / 2;
-      dom.scrollTo({top: (dom.clientHeight / 24) * middleHour, behavior: 'smooth'});
+      if (this.calendar.getViewName() !== 'month') {
+        //Scoll to the middle of the event, if it's not month view
```
paulfitz, line 92: "//Scoll" -> "// Scroll"

What is this trying to solve? (add answer as a comment in code please)

calendar/page.js
```diff
@@ -82,31 +88,38 @@ class CalendarHandler {
...
   // change calendar perspective between week, month and day.
   changeView(calendarViewPerspective) {
     this.calendar.changeView(calendarViewPerspective);
+    document.getElementById('calendar-title').innerText = getMonthName();
   }
 
   // navigate to the previous time period
   calendarPrevious() {
     this.calendar.prev();
+    setMonthNameInDom();
```
paulfitz, line 110: Perhaps a more general `updateAfterNavigation` name would work, could imagine needing to adapt the UI in several ways.

calendar/page.js
```diff
@@ -82,31 +88,38 @@ class CalendarHandler {
...
   calendarNext() {
     this.calendar.next();
+    setMonthNameInDom();
   }
 
   //navigate to today
   calendarToday() {
     this.calendar.today();
+    setMonthNameInDom();
   }
```
paulfitz, line 124: Aren't there going to be other ways the period could change? E.g. via linking? Trying out linking locally, I can jump between months without the month name being updated.

JakubSerafin: Yes, it was a case that I've missed

calendar/page.js
JakubSerafin: > what is .d.d?

TUI Calendar is using custom TZDate structure to handle the dateTime and  i used [..].d.d.[...] to access plain old Date format. Nerveless i found more canonical way to do this is documentation :)

JakubSerafin:
> Test failure is replicable with MOCHA_WEBDRIVER_HEADLESS=1

even with a flag on i was not able to reproduce error on my machine, despite running tests multiple time. It's disturbing

calendar/page.js
```diff
@@ -88,8 +88,11 @@ class CalendarHandler {
       this.calendar.updateEvent(record.id, CALENDAR_NAME, {backgroundColor: CalendarHandler._selectedColor});
       this._selectedRecordId = record.id;
       this.calendar.setDate(record.startDate);
+      updateUIAfterNavigation();
       if (this.calendar.getViewName() !== 'month') {
-        //Scoll to the middle of the event, if it's not month view
+        //Scroll to the middle of the event if it's not month view.
```
paulfitz, line 93: Thanks! Could you please add a space after the `//`s, to match the bulk of Grist source code.

calendar/page.js
```diff
@@ -194,7 +197,8 @@ function getGristOptions() {
 }
 
 
-function setMonthNameInDom(){
+function updateUIAfterNavigation(){
+  //update name of the month and year displaed on the top of the widget
```
paulfitz, line 201: displaed -> displayed
[also spaces after // would save me time deciding whether it is a big enough deal to mention :)]

calendar/page.js
```diff
@@ -82,31 +88,41 @@ class CalendarHandler {
...
+          + (record.endDate.getHours() - record.startDate.getHours()) / 2;
+        dom.scrollTo({top: (dom.clientHeight / 24) * middleHour, behavior: 'smooth'});
+      }
     }
   }
 
   // change calendar perspective between week, month and day.
   changeView(calendarViewPerspective) {
     this.calendar.changeView(calendarViewPerspective);
+    document.getElementById('calendar-title').innerText = getMonthName();
```
paulfitz, line 107: Could this be a call to `updateUIAfterNavigation`

paulfitz (commented):
A few very small comments, pretty much there.

I can't replicate test failure on latest version, whereas 2acd12c failed consistently for me in headless.

calendar/page.js
JakubSerafin: In current moment there is no point - changing perspective should persist day that calendar is focused on, so it should not change neither month nor year that is showed in header. But in future other UI element can be updated after navigation, so I'll add it

paulfitz: I'm saying that if you think of `updateUIAfterNavigation` as being in charge of updating the UI after some significant change, there's no need to duplicate the `document.getElementById('calendar-title').innerText = getMonthName();` code here - that can be the responsibility of `updateUIAfterNavigation`.

paulfitz:
There's now a "changes you made may not be saved" style alert when trying to switch to another test.

Screenshot from 2023-09-12 09-17-21

paulfitz:
@JakubSerafin I looked at what grist-core does and moved some code (https://github.com/gristlabs/grist-core/pull/677).

I see that this kind of alert is dismissed during simulated logins when working with an external server, which is like our situation in grist-widget:
https://github.com/gristlabs/grist-core/blob/bc6f5504712e47c19ab6b3e9244affc508d63812/test/nbrowser/homeUtil.ts#L115
I'd be interested in knowing what this is about, but it doesn't seem specifically a grist-widget issue - and commenting out calendar tests entirely, I can still see failures due to it.

So I just added a handler, like you did but at a wider scope.

paulfitz:
@JakubSerafin btw I noticed when editing it that spacing is off in test/calendar.ts (4 versus 2)

PR #60, opened 2023-09-08
Update flashcards interface colors to match the Grist product

dsagal:


flashcards/index.html
```diff
@@ -13,6 +13,15 @@
         margin: 0;
         padding: 0;
         font-family: sans-serif;
+
+        --color-button-primary-fg: white;
```
berhalak, line 17: Tiny suggestion: I'd go with `:root` selector, now those variables are defined "twice" in the dom.

berhalak (approved):
Thanks.

PR #61, opened 2023-09-08
Reset progress when switching to a different set

dsagal:


berhalak (approved):
Works well, thank you.

PR #63, opened 2023-09-09
Add link to Notepad widget toolbar

georgegevoian:


notepad/index.js
```diff
@@ -20,7 +20,9 @@ const toolbarOptions = [
   [{ 'font': [] }],
   [{ 'align': [] }],
 
-  ['clean']                                         // remove formatting button
+  ['clean'],                                        // remove formatting button
+
+  ['link'],
```
berhalak, line 25: Can you change the order. The clear should by last in my opinion.

berhalak (approved):
Thanks, one suggestion only.

PR #64, opened 2023-09-18
Polish calendar widget

georgegevoian:
- clicking or creating an event now moves focus to the widget
 - new events can no longer be created if the widget isn't configured
 - isAllDay can be set even when there is no endDate
 - events that start and end at 12:00am are now visible in all views
 - an invalid or missing endDate is now handled by using one that is 30 minutes after startDate

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -21,15 +21,12 @@ function ready(fn) {
 }
 
 function isRecordValid(record) {
-  return ( 
-    record.startDate instanceof Date &&
-    (
-      (record.endDate === null && typeof record.isAllDay === 'boolean') ||
-      record.endDate instanceof Date
-    ) &&
-    typeof record.title === 'string' &&
-    (record.isAllDay === undefined || typeof record.isAllDay === 'boolean')
-  );
+  const hasStartDate = record.startDate instanceof Date;
+  const hasTitle = typeof record.title === 'string';
+  const hasEndDateOrIsAllDay = record.endDate instanceof Date ||
+    (record.endDate === null && typeof record.isAllDay === 'boolean');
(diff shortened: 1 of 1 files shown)


JakubSerafin:
@georgegevoian  - I've added one line that make widget focus when any event on calendar is moved/resized. Hope you don't mind

calendar/page.js
```diff
@@ -21,9 +21,15 @@ function ready(fn) {
 }
 
 function isRecordValid(record) {
-  return record.startDate instanceof Date &&
-  record.endDate instanceof Date &&
-  typeof record.title === 'string'
+  return (
```
JakubSerafin, line 24: maybe 
```

function isRecordValid(record) {
  const haveAStartDate = () => record.startDate instanceof Date;
  const haveATitle = () => typeof record.title === 'string';
  const haveAnEndDateOrIsAllDay = () => (record.endDate === null && typeof record.isAllDay === 'boolean') ||
  record.endDate instanceof Date;
  const allDayIsValidFormat = () => record.isAllDay === undefined || typeof record.isAllDay === 'boolean';
  
  return haveAStartDate() && haveATitle() && haveAnEndDateOrIsAllDay() && allDayIsValidFormat();
}
```

?

calendar/page.js
```diff
@@ -310,6 +332,14 @@ async function updateCalendar(records, mappings) {
   dataVersion = Date.now();
 }
 
+function defaultFocus() {
```
JakubSerafin, line 335: why "default"Focus? 
Form what i can see, this method just put focus on the window, so why not "focusOnWidget"?

JakubSerafin (approved):
Thanks!

PR #65, opened 2023-09-18
Feature/translation for calendar widget

JakubSerafin:
Calendar Widget translation provided by i18next. Translation files are stored in ..\i18n\[language]\translation.json. Right now it's only polish. 
When no translation is found, text used as a key is an fallback (just like in grist approach to translation) 

New test is added to check if widget is translating itself when language is changed, but it would fail until grist repository won't be merged with changes to send language information to custom widget.


calendar/page.js
```diff
@@ -194,7 +233,35 @@ async function configureGristSettings() {
...
+
+async function translatePage() {
+
+  const backendOptions = {
+
+    loadPath: 'i18n/{{lng}}/{{ns}}.json',
+    addPath: 'i18n/add/{{lng}}/{{ns}}',
+    // allow cross domain requests
+    crossDomain: false,
+    // allow credentials on cross domain requests
```
georgegevoian, line 249:
```suggestion
    // don't allow cross domain requests
    crossDomain: false,
    // don't include credentials on cross domain requests
```

test/calendar.ts
```diff
@@ -46,118 +46,181 @@ describe('calendar', function () {
         await grist.setCustomWidgetMapping('endDate', /To/);
         await grist.setCustomWidgetMapping('title', /Label/);
         await grist.setCustomWidgetMapping('isAllDay', /IsFullDay/);
+        //sing in to grist
```
georgegevoian, line 49:
```suggestion
        //sign in to grist
```

test/calendar.ts
```diff
@@ -46,118 +46,181 @@ describe('calendar', function () {
...
 
-            const newDate = await grist.executeScriptOnCustomWidget<string>(
-              'return calendarHandler.calendar.getDate().d.toDate().toDateString()'
-            );
+    await selectPerspective('month');
+    await navigateCalendar('today');
 
-            const expectedDate = new Date(today);
-            expectedDate.setDate(today.getDate() + daysToAdd);
+    // TUI Calendar use it's own data enegine, so mocking Date is not working here. In future it will be good to
```
georgegevoian, line 167:
```suggestion
    // TUI Calendar use it's own data engine, so mocking Date is not working here. In future it will be good to
```

test/getGrist.ts
```diff
@@ -23,6 +23,13 @@ export function getGrist(): GristUtils {
     // TODO: mocha-webdriver has a way of explicitly connecting a
     // server that might have advantages for debugging.
     await grist.wait();
+    // "Changes you made may not be saved" alerts are consumed
+    // in grist-core tests - do the same here.
+    // TODO: figure out why they occur.
```
georgegevoian, line 28: If it helps: they now appear in Grist when navigating away from the page while the current document is unsaved. (Many tests create a temporary, unsaved document.)

This was added recently.

test/getGrist.ts
```diff
@@ -188,6 +197,12 @@ export class GristUtils extends GristWebDriverUtils {
     await this.waitForServer();
   }
 
+  public async closeDoc(): Promise<void> {
+    await driver.get(this.url);
+    await this.waitForServer();
```
georgegevoian, line 202: Does this line do anything? The line above it navigates to a URL in the browser, but doesn't wait for the page to load, so I'm not sure `waitForServer()` will actually be able to monitor network activity yet.

PR #68, opened 2023-09-25
Enable keyboard shortcuts in calendar

georgegevoian:


calendar/page.js
```diff
@@ -220,6 +220,8 @@ async function configureGristSettings() {
   // When options changed in the widget configuration (reaction to perspective change)
   grist.onOptions(onGristSettingsChanged);
 
+  grist.enableKeyboardShortcuts();
```
berhalak, line 223: Calendar widget loads this file from the `main` source, so this PR will break this widget for a while. So maybe some coordination with release, or some conditional here should be added.

georgegevoian: Any preference? I was thinking of assigning a task for myself to monitor the release Sunday night and merge this after it succeeds.

paulfitz: I'd vote for `enableKeyboardShortcuts?.()` for now so that you don't have to do heroics @georgegevoian.

JakubSerafin: Maybe conditionally calling this method if it exist with TODO to remove condition in latter term?

berhalak (approved):
Thanks.

PR #69, opened 2023-09-25
support for creating widget bundles

paulfitz:
It can be useful to include custom widgets with Grist, so they can be used offline, or for simpler archiving, or for tighter integration with the rest of the app. This PR includes a script for collecting custom widget files in a structured bundle. It is parallel work to developments in grist-core, including https://github.com/gristlabs/grist-core/pull/714

### DIFF
--- a/buildtools/bundle.js
+++ b/buildtools/bundle.js
@@ -1,0 +1,242 @@
+#!/usr/bin/env node
+
+/**
+ *
+ * This is a script to bundle up some widgets and their
+ * assets into a standalone directory that could be used
+ * offline. It is a little bit string and duct-tape-y,
+ * sorry about that.
+ *
+ * Requires wget to operate. Will only bundle widgets
+ * that have an "archive" field in their package.json.
+ * The "archive" field can be an empty {}, or:
+ *    {
+ *      "domains": ["domain1.com", "domain2.com"],
+ *      "entrypoints: ["https://other.page"]
+ *    }
+ * Material in the specified domains will also be
+ * included in the bundle. Material needed by any
+ * extra entry-point pages will be included as well.
+ *
+ * The bundle will also include grist-plugin-api.js.
+ * This is not intended for use, but rather as a
+ * placeholder for where Grist could serve its own
+ * version of this file.
+ *
+ * The bundle includes a manifest.json file with
+ * relative URLs to the included widgets.
+ *
+ * The bundle also includes a manifest.yml file to
+ * describe the widgets as a plugin to Grist. The
+ * clash in file names is a little unfortunate.
+ *
+ * Call without any arguments. Run from the root of the
+ * repository. Places results in:
+ *    dist/grist-widget-bundle
+ *
+ * Will run a temporary server on port 9990.
+ *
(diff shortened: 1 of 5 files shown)


buildtools/bundle.js
```diff
@@ -0,0 +1,242 @@
...
+    this.host = `localhost:${this.port}`;  // Host for serving widgets.
+    this.renamedHost = 'widgets';          // Final directory name for widgets.
+    this.assetUrl = `http://${this.host}`; // URL for serving widgets.
+    this.widgets = [];         // Bundled widgets will be added here.
+    this.bundledAt = null;     // A single time that applies to whole bundle.
+    this.bundleRoot = null;    // Where bundle will be stored.
+  }
+
+  // Start a widget server.
+  start() {
```
berhalak, line 72: Maybe something like this, to make sure it works, (I have python 2.7 by default so it breaks for me):
```
  async start() {
    this.localServer = spawn('python', [
      '-u', '-m', 'http.server', this.port,
    ], {
      cwd: process.cwd(),
    });

    await new Promise((resolve, reject) => {
      this.localServer.stdout.on('data', (data) => {
        if (data.toString().includes('Serving HTTP on')) {
          resolve();
        }
      });
      this.localServer.stderr.on('data', (data) => {
        reject(data.toString());
      });
    });
  }
  
```

buildtools/bundle.js
```diff
@@ -0,0 +1,242 @@
...
+ * The "archive" field can be an empty {}, or:
+ *    {
+ *      "domains": ["domain1.com", "domain2.com"],
+ *      "entrypoints: ["https://other.page"]
+ *    }
+ * Material in the specified domains will also be
+ * included in the bundle. Material needed by any
+ * extra entry-point pages will be included as well.
+ *
+ * The bundle will also include grist-plugin-api.js.
```
berhalak, line 21: This doesn't work for me, should it be added to the domains?

berhalak (commented):
I think dist can be added to .gitignore.

buildtools/bundle.js
paulfitz: Is this to wait for the server to be up and responsive, instead of the `wait()` method? I think I'd still have a weak preference for waiting for a `fetch()` to succeed, since it seems easier to be confident of it. When checking a message on console, I find myself asking is this printed before or after the server is started. I'm a bit daunted about tracking down exact python source code, but it may be like this:
https://github.com/python/cpython/blob/20cfab903db70cf952128bc6b606e3ec4a216498/Lib/http/server.py#L1267-L1275
which at first glance looks like it prints the message first.

Separately, maybe I should write `python3` instead of `python`?

buildtools/bundle.js
paulfitz: Ah, sorry, I had commented out a `// domains.push('getgrist.com')` line. I put it back in.

buildtools/bundle.js
berhalak: Yes, python3 will be enough :)

berhalak (approved):
Thanks.

PR #74, opened 2023-09-27
Feature/dot in month view has accent color when selected

JakubSerafin:
When calendar is in month view, and row is selected, "dot" for single day event is displayed in accent color.


georgegevoian (approved):
Works great. Thanks!

PR #75, opened 2023-09-27
Support Date columns and fix click-to-drag bug (WIP)

georgegevoian:


calendar/page.js
```diff
@@ -472,6 +505,10 @@ function thirtyMinutesFrom(date) {
   return new Date(date.getTime() + 30 * 60 * 1000);
 }
 
+function hasNonzeroTime(date) {
```
berhalak, line 508: There timezones that are shifted by minutes, so this won't work in all cases, but I think this is good enough for now.

JakubSerafin:
@georgegevoian  It Is still Work In Progress? This code is used a a part of https://github.com/gristlabs/grist-widget/pull/84 and it would be nice to merge it to master

dsagal:
Actually #84 does not rely on this landing, it incorporated these changes already, and also changed the logic to use column types instead, so the code is incompatible.

PR #76, opened 2023-09-27
Fix bugs with new read-only mode

georgegevoian:


dsagal (approved):
Looks great, thank you!

PR #78, opened 2023-09-27
[Calendar] Fixing theme variables and styles for selected card

berhalak:
Fixing typos in theme properties which is fixing the style for the selected card.


georgegevoian (approved):
Thanks @berhalak!

PR #79, opened 2023-09-27
Use new icons for arrow buttons

georgegevoian:


paulfitz (approved):
Looks good!

calendar/screen.css
```diff
@@ -73,6 +107,13 @@ body {
   border-color: #009058;
 }
 
+.calendar-button-previous,
```
berhalak, line 110: This doesn't work. Those are identifiers not classes in HTML.

PR #82, opened 2023-09-28
Make calendar endDate optional

georgegevoian:


calendar/page.js
```diff
@@ -26,11 +26,10 @@ function ready(fn) {
 
 function isRecordValid(record) {
   const hasStartDate = record.startDate instanceof Date;
+  const maybeHasEndDate = !record.endDate || record.endDate instanceof Date;
```
berhalak, line 29: Why ! ?

berhalak (changes requested):

calendar/page.js
berhalak: Ahh, now I get it. It was hard to understand.

berhalak (approved):
Thanks.

PR #87, opened 2023-09-28
With Date column type, be more careful about timezones for calculating the Grist end date

dsagal:


berhalak (approved):
Works for me, thanks.

PR #90, opened 2023-09-28
Add minimal accent to Sunday in calendar

georgegevoian:


calendar/screen.css
```diff
@@ -333,6 +333,11 @@ input:checked + label {
   padding: 0px;
 }
 
+.toastui-calendar-week-view .toastui-calendar-holiday-sun,
```
JakubSerafin, line 336: why not use theme property instead: 
https://github.com/nhn/tui.calendar/blob/main/docs/en/apis/theme.md#common-holiday ?

JakubSerafin (changes requested):

calendar/screen.css
georgegevoian: It applies the accent to more places than these changes, which isn't desirable (e.g. the month view).

dsagal (approved):
It looks fine, thank you. Other cleanup could follow later.

PR #92, opened 2023-09-28
Fix selectRecord for Date columns

georgegevoian:


calendar/page.js
```diff
@@ -509,13 +509,30 @@ function selectRadioButton(value) {
...
 
+/**
+ * Returns a new date that's shifted towards UTC+0 if `colType` is `Date`.
+ * 
+ * Returns `date` unchanged if `colType` is `DateTime`.
+ */
+function getAdjustedDate(date, colType) {
+  if (colType !== 'Date') { return date; }
+
+  const ms = date.valueOf() + (date.getTimezoneOffset() * 60000);
```
dsagal, line 520: You could add a comment that this is similar to `date.tz('UTC')` but not broken when there is a DST difference between now and date.

dsagal (approved):
Thank you!

PR #93, opened 2023-09-29
Optimize calendar performance and fix bugs

georgegevoian:


calendar/page.js
```diff
@@ -304,31 +315,50 @@ class CalendarHandler {
...
+    const newVisibleEventIds = new Set();
+    const dateRangeStart = this.calendar.getDateRangeStart();
+    const dateRangeEnd = this.calendar.getDateRangeEnd().setHours(23, 99, 99, 999);
+
+    // Add or update events that are now visible.
+    for (const event of this._allEvents.values()) {
+      const isEventInRange = (
+        (event.start >= dateRangeStart && event.start <= dateRangeEnd) ||
+        (event.end >= dateRangeStart && event.end <= dateRangeEnd) ||
+        (event.start >= dateRangeStart && event.start <= dateRangeEnd) ||
```
paulfitz, line 340: hang on what is this third one?

calendar/page.js
```diff
@@ -304,31 +315,50 @@ class CalendarHandler {
...
+        (event.start < dateRangeStart && event.end > dateRangeEnd)
+      );
+      if (!isEventInRange) { continue; }
+  
+      const calendarEvent = this.calendar.getEvent(event.id, CALENDAR_NAME);
+      if (!calendarEvent) {
+        this.calendar.createEvents([event]);
```
paulfitz, line 347: I think you said you checked that bundling for a single call to createEvents has negligible benefit, am I remembering that right? Otherwise I'd be suggesting looking into it.

test/calendar.ts
```diff
@@ -20,14 +15,26 @@ describe('calendar', function () {
...
   }
 
-  //wait until the event is loaded on the calendar
+  // Not a pretty way to get events from currently used calendar control. but it's working.
+  function buildGetVisibleCalendarEvent(eventId: number) {
+    return `return testGetVisibleCalendarEvent(${eventId});`
+  }
+
+  async function getVisibleCalendarEvent(eventId: number): Promise<any> {
+    const event = await grist.executeScriptOnCustomWidget(buildGetVisibleCalendarEvent(eventId));
```
paulfitz, line 24: Somehow feels like whoever wrote `executeScriptOnCustomWidget` went against the grain, forcing functions to be expressed as strings. This is uglier than any custom widget testing in grist-core.

calendar/page.js
georgegevoian: Looks like a duplicate of the first. Thanks for catching that.

calendar/page.js
georgegevoian: Yeah, I tried again recently with a large number of events and benchmarked it. Didn't see any notable improvements and the code looks a bit cleaner without it.

calendar/page.js
```diff
@@ -2,18 +2,16 @@
...
 const isReadOnly = urlParams.get('readonly') === 'true' ||
   (urlParams.has('access') && urlParams.get('access') !== 'full');
 
-//for tests
-let dataVersion = Date.now();
-function testGetDataVersion(){
-  return dataVersion;
-}
+// for tests
+var dataVersion = Date.now();
```
paulfitz, line 14: Why are we using var now?

georgegevoian: It gets added to `window` automatically. But I should also be able to do `window.dataVersion = dataVersion;` if that looks better.

.github/workflows/main.yml
georgegevoian, line 1: Just syncing the Node version with the main Grist app here.

package.json
```diff
@@ -12,26 +12,22 @@
     "test": "docker image inspect gristlabs/grist --format 'gristlabs/grist image present' && NODE_PATH=_build SELENIUM_BROWSER=chrome mocha -g \"${GREP_TESTS}\" _build/test/*.js",
     "test:ci": "MOCHA_WEBDRIVER_HEADLESS=1 npm run test",
     "pretest": "docker pull gristlabs/grist && tsc --build && node ./buildtools/publish.js http://localhost:9998",
-    "grist": "docker run -t -i --rm --name grist-dev --network=host -e PORT=8484 -e GRIST_SINGLE_ORG=preview -e GRIST_WIDGET_LIST_URL=http://localhost:8585/manifest.json gristlabs/grist"
```
georgegevoian, line 15: Removing `--network=host` here and elsewhere because it's unsupported on macOS.

paulfitz: What is this supposed to do? I don't see any exposed port to connect to? Also what is host-gateway used for?

Once I figure out how this is supposed to be used, I'd like to double check it, since in the past host.docker.internal worked differently on macs vs linux (hopefully that's all fixed long ago, but I'd like to double check)

test/getGrist.ts
```diff
@@ -63,10 +63,10 @@ export class GristTestServer {
     await this.stop();
     const {gristContainerName, gristImage, gristPort, contentPort} = serverSettings;
     const cmd = `docker run -d --rm --name ${gristContainerName}` +
-      ` --network="host"` +
+      ' --add-host=host.docker.internal:host-gateway' +
```
paulfitz, line 66: What role does host-gateway play?

test/getGrist.ts
```diff
@@ -117,7 +117,8 @@ export class GristTestServer {
 
   public get assetUrl() {
     const {contentPort} = serverSettings;
-    return `http://localhost:${contentPort}`;
+    // localhost doesn't work on Node 18 (see https://github.com/node-fetch/node-fetch/issues/1624)
```
paulfitz, line 120: That's a bit scary. I see it is ipv6 related - https://github.com/node-fetch/node-fetch/issues/1624#issuecomment-1407717012

Does this have any consequences for the move to node 18 in Grist? I feel like there's a lot of dependence on localhost in test machinery?

georgegevoian: I don't think we make any outbound requests to `localhost` from Node, and the failures I've seen since the upgrade feel more symptomatic of general flakiness.

I see one option is to do `dns.setDefaultResultOrder('ipv4first')` to go back to the pre-17 behavior.

georgegevoian: Yeah. I suspect that all may still be working. Even in this file, the fetch to localhost works fine for the Grist instance running in Docker - it's just live-server that is having the ipv6 problem. We should take a closer look after.

test/getGrist.ts
georgegevoian: I don't have a Linux machine to test with (though I will go ahead and remove the flag and re-run the CI to check), but I believe this is required for `host.docker.internal` to work on Linux.

package.json
georgegevoian: That's a good point - I'm looking now and this script looks unused. May be ok to remove?

test/getGrist.ts
georgegevoian: What do you think about checking the OS before building the docker command, and only applying these changes if it's not Linux? If I understand the situation right on Windows, it has the same limitation as macOS.

paulfitz (approved):
Read about the host-gateway thing and I see what it does on newer docker :+1:

package.json
berhalak: I think I added it, so that it is easy to just build widgets and test them manually. Fine with removing it, or just leaving it as it is for linux users :).
(as this uses host network, docker container is just available on localhost:8484, there is no network isolation, so no ports forwarding)

PR #94, opened 2023-09-29
events show full text when the cursor is on hover above its name

JakubSerafin:
According to the work item: 
https://gristlabs.getgrist.com/doc/check-ins/p/5#a1.s9.r2811.c18

Now events have a full text displayed if one hover cursor over the event texts.
obraz


georgegevoian (approved):
Thanks!

PR #99, opened 2023-10-04
Add widgets for custom charts and jupyterlite notebooks

alexmojaki:


chart/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@gristlabs/widget-chart",
+  "description": "Widget for more customizable charts with plotly",
+  "homePage": "https://github.com/gristlabs/custom-charts-widget",
+  "version": "0.0.1",
+  "grist": [
+    {
+      "name": "Chart",
```
paulfitz, line 8: A more specific name would be better.

paulfitz, line 8: A more specific name would be better.

jupyterlite/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@gristlabs/widget-jupyterlite",
+  "description": "Widget for a JupyterLite notebook",
+  "homePage": "https://github.com/gristlabs/jupyterlite-widget",
+  "version": "0.0.1",
+  "grist": [
+    {
+      "name": "Chart",
```
paulfitz, line 8: The name is out of sync

paulfitz (changes requested):
This approach worries me, and changes how the widget collection works in ways that could be hard to step back from. I'd rather have someone look seriously at alternatives before committing. Placing this on hold, since use of these widgets by URL is enough for now.

berhalak:
How about small mental model shift. Lets treat this github repository as a package repository (like npm), so it contains all prebuilt artifacts (js files). And in the chart and jupiter repository, you would just add a github action that will build your project and create a pull request for this one, with everything built already and configured properly?
So you will use it (this repo) as a hosting provider for your widgets, and all the source code and tests and all the machinery will still sit in your repository.

paulfitz (approved):
Thanks @alexmojaki !

PR #109, opened 2023-11-14
create a widget bundle for insertion into grist-core via npm packaging

paulfitz:
This adds a few more flags to the bundling script, and then uses them in a `prepare` script to assemble a bundle for insertion into grist-core. The bundle can be previewed as a tarball with:
```
yarn pack
```
or published to the npm registry as [@gristlabs/grist-widget](https://www.npmjs.com/package/@gristlabs/grist-widget) with:
```
yarn publish --access public
```

There will be a companion PR to grist-core to make use of the package.

### DIFF
--- a/buildtools/bundle.js
+++ b/buildtools/bundle.js
@@ -30,9 +30,14 @@
  * describe the widgets as a plugin to Grist. The
  * clash in file names is a little unfortunate.
  *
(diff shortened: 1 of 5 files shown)


georgegevoian (approved):
Thanks @paulfitz.

PR #110, opened 2023-11-14
Fix Invoice and Purchase Order widgets to make text visible in dark mode

dsagal:
CSS tweaks. Light mode looks the same as before the change.

Dark mode before:
Screenshot 2023-11-14 at 12 06 51 PM
Screenshot 2023-11-14 at 12 07 11 PM

Dark mode after:
Screenshot 2023-11-14 at 12 06 15 PM
Screenshot 2023-11-14 at 12 06 26 PM

### DIFF
--- a/invoices/invoice.css
+++ b/invoices/invoice.css
@@ -53,7 +53,7 @@ body {
 
 .client .details, .summary .details {
   padding: 4px;
-  background-color: var(--grist-theme-page-panels-main-panel-bg);
+  background-color: var(--grist-theme-page-panels-main-panel-bg, white);
 }
 
 table.items {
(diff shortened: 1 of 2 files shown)


georgegevoian (approved):
Thanks!

invoices/invoice.css
```diff
@@ -54,7 +53,7 @@ body {
 
 .client .details, .summary .details {
   padding: 4px;
-  background: #fff;
+  background-color: var(--grist-theme-page-panels-main-panel-bg);
```
georgegevoian, line 56: I think we need to add fallbacks for these. (The variables don't always exist, like when custom CSS is enabled.)

georgegevoian (changes requested):

invoices/invoice.css
dsagal: Thanks. Done.

georgegevoian (approved):
Thanks @dsagal!

PR #114, opened 2023-11-28
update Calendar test

paulfitz:
There is an extra section due to the new Record Card feature, so ids have changed. (thanks for tip @georgegevoian)


georgegevoian (approved):
Thanks @paulfitz!

PR #116, opened 2023-12-13
Add chart submodule

alexmojaki:


JakubSerafin (approved):

alexmojaki:
Why doesn't https://deploy-preview-116--boisterous-sunburst-a5c941.netlify.app/manifest.json work?

Also it's very cool how the preview link above automatically ends in `/chart`. Had me worried at first that something was wrong.

PR #119, opened 2023-12-15
Add test for chart widget

alexmojaki:


paulfitz (approved):
Super, thanks @alexmojaki

PR #122, opened 2024-01-16
Open Record Card on Calendar event click

georgegevoian:
New test will fail temporarily until the Grist image is updated with support for opening a Record Card from a custom widget.

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -218,6 +218,7 @@ class CalendarHandler {
     TZDate = this.calendar.getDate().constructor;
 
     this.calendar.on('clickEvent', async (info) => {
+      focusWidget();
       await grist.setCursorPos({rowId: info.event.id});
     });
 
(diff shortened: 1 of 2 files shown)


calendar/page.js
```diff
@@ -218,6 +218,7 @@ class CalendarHandler {
     TZDate = this.calendar.getDate().constructor;
 
     this.calendar.on('clickEvent', async (info) => {
+      focusWidget();
```
berhalak, line 221: Why this is needed?

berhalak (approved):
Thanks, one question only.

calendar/page.js
georgegevoian: Without this, the calendar widget doesn't get focus if you click or double-click an event. This breaks the new `viewAsCard` command because it opens the selected record in the active section, and if the calendar isn't the active section, the wrong record is shown in the popup.

PR #123, opened 2024-01-30
Save Markdown on blur and tweak CSS

georgegevoian:


berhalak (changes requested):
Now it is not possible to cancel, am I right? It feels buggy now, rather then how it worked before, I can by accident delete everything. I've read the community post, and the problem was that we don't offer auto-save feature, not that this widget had bugs or any issues.

From my perspective, we should offer the "auto-save" feature, and then render button called "Preview"/"Editor", where people will just switch.
WDYT?

georgegevoian:
> Now it is not possible to cancel, am I right? It feels buggy now, rather then how it worked before, I can by accident delete everything. I've read the community post, and the problem was that we don't offer auto-save feature, not that this widget had bugs or any issues.
> 
> From my perspective, we should offer the "auto-save" feature, and then render button called "Preview"/"Editor", where people will just switch. WDYT?

Talked it through with Anais and we decided to add a secondary cancel button. How does that sound?

berhalak (approved):
Thanks.
Do you think it is worth centering it verticaly ?
image

PR #124, opened 2024-02-15
Allow several action buttons

jperon:
Original idea of @emanuelegissi

If, instead of a dictionary, a list of dictionaries is defined, several buttons will be displayed.

This is backward-compatible: if only one dictionary is defined, it will work as before (except for styling changes).

See in action here: https://docs.getgrist.com/dd1Amj8bbHzK/Tests/p/23#a1.s135.r1.c219.

### DIFF
--- a/actionbutton/actionbutton.js
+++ b/actionbutton/actionbutton.js
@@ -45,7 +45,7 @@ function onRecord(row, mappings) {
     }
     let btns = row[column]
     // If only one action button is defined, put it within an Array
-    if (!btns.length) {
+    if (!Array.isArray(btns)) {
       btns = [ btns ]
     }
     const keys = ['button', 'description', 'actions'];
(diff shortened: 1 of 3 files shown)


actionbutton/actionbutton.js
```diff
@@ -42,15 +43,22 @@ function onRecord(row, mappings) {
     if (!row.hasOwnProperty(column)) {
       throw new Error(`Need a visible column named "${column}". You can map a custom column in the Creator Panel.`);
     }
+    let btns = row[column]
+    // If only one action button is defined, put it within an Array
+    if (!btns.length) {
```
berhalak, line 48: Can you add a comment here that here we actually are testing if the `length` is in the object, not actually the length of an array. Or maybe it is better to use `Array.isArray` method.

berhalak (approved):
Thanks @jperon!

PR #125, opened 2024-03-11
Migration tool from Spreadsheet.com to Grist

dsagal:
There is no need to merge this diff in. I am making a PR to get a preview up. The preview should be good enough to have a working template. This may be useful for up to 3 more months, but does not need to be available long-term, and does not need to show up as a choice in people's documents.


paulfitz:
> This may be useful for up to 3 more months

ok to close now, do you think @dsagal ?

dsagal:
Yes, closing. Thanks for the reminder.

PR #129, opened 2024-06-21
Add optional HTML export to Notepad widget

### LINKED ISSUE #128: Notepad widget should allow HTML export to another column
Quill/delta format is not standard, therefore it is hard to reuse in other contexts.

Exporting directly to HTML would allow easy read to the already existing easy write feature allowed by the notepad widget.

guillett:
Hi,

My attempt to fix #128 .

Working deployment on https://guillett.github.io/grist-widget/ (notepad only)

### DIFF
--- a/notepad/index.html
+++ b/notepad/index.html
@@ -5,9 +5,9 @@
   <meta charset="utf-8">
   <title>Notepad</title>
   <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
-  <link href="//cdn.quilljs.com/1.3.6/quill.snow.css" rel="stylesheet">
-  <link href="//cdn.quilljs.com/1.3.6/quill.bubble.css" rel="stylesheet">
-  <script src="//cdn.quilljs.com/1.3.6/quill.min.js"></script>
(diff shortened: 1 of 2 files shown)


notepad/index.html
```diff
@@ -5,9 +5,9 @@
   <meta charset="utf-8">
   <title>Notepad</title>
   <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
-  <link href="//cdn.quilljs.com/1.3.6/quill.snow.css" rel="stylesheet">
-  <link href="//cdn.quilljs.com/1.3.6/quill.bubble.css" rel="stylesheet">
-  <script src="//cdn.quilljs.com/1.3.6/quill.min.js"></script>
+  <link href="https://cdn.jsdelivr.net/npm/quill@2.0.2/dist/quill.snow.css" rel="stylesheet">
```
berhalak, line 8: Do we need to update the editor? The major version has changed, so it might work differently.

guillett: Unfortunately, `getSemanticHTML` is not available in `v1`.

berhalak (approved):
Thanks. I checked release notes and it looks fine. No breaking changes.

PR #135, opened 2024-08-18
Update chart and jupyterlite

georgegevoian:


paulfitz (approved):
Thanks @georgegevoian

PR #139, opened 2024-09-21
Fixes Calendar not working offline

Spoffy:
The calendar widget currently doesn't work in Grist Desktop, due to it trying to fetch certain dependencies from the internet.

This PR adds those dependencies to the bundle, enabling offline use.

## Testing

This change was tested locally by producing a new bundle and rebuilding Grist Desktop.

### DIFF
--- a/calendar/package.json
+++ b/calendar/package.json
@@ -12,7 +12,7 @@
       "accessLevel": "full",
       "renderAfterReady": true,
       "archive": {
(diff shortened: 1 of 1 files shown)


berhalak:
Do you know why this happened? It used to work offline.

berhalak (approved):
Ahh, we added translations. Can you add an issue (or a task) to add some kind of safe check if all external links were resolved?

Spoffy:
Added #140 on the issue tracker to cover this.

PR #144, opened 2024-10-08
Fix tests broken due to widget gallery changes

Spoffy:


paulfitz:
Took a quick look at this PR. A few notes.

`test/gristWebDriverUtils.ts` has become weird. It is supposed to be a copy of an asset in `grist-core`:

```
/**
 * Utilities that simplify writing browser tests against Grist, which
 * have only mocha-webdriver as a code dependency. Separated out to
 * make easier to borrow for grist-widget repo.
 *
 * If you are seeing this code outside the grist-core repo, please don't
 * edit it, it is just a copy and local changes will prevent updating it
 * easily.
 */
```

It looks pretty out of date with `grist-core` today, so I thought about copying over the latest version. BUT it looks like there have been local changes to the file. For example the `focusOnCell` method. These should either be added upstream, or in `GristUtils` in `test/getGrist.ts`.

Copying a file across is of course lame.

test/getGrist.ts
```diff
@@ -17,6 +17,8 @@ export function getGrist(): GristUtils {
   const server = new GristTestServer();
   const grist = new GristUtils(server);
 
+  enableDebugCapture()
```
dsagal, line 20: Missing semicolon.

test/gristWebDriverUtils.ts
```diff
@@ -210,8 +223,11 @@ export class GristWebDriverUtils {
     const max = 10;
 
     // Keep dismissing prompts until there are no more, up to a maximum of 10 times.
-    while (i < max && await this.driver.find('.test-behavioral-prompt').isPresent()) {
-      await this.driver.find('.test-behavioral-prompt-dismiss').click();
+    const getButton = () => {
```
dsagal, line 226: My guess is that this change isn't of much benefit, and my preference would be to keep it unchanged, since it's a copy of code from grist-core (which is much more exercised and reliable). For the same reason, `ignoreMissingElementErrors` doesn't seem needed.

dsagal (commented):
Thank you! One note about a change which I suspect is not needed to get the fixes of this PR.

test/gristWebDriverUtils.ts
Spoffy: Rolled this back - totally agree it doesn't make sense making these changes, if we haven't updated gristWebDriverUtils.

dsagal (approved):

PR #145, opened 2024-11-20
editor: Fixing bug that caused infinite loop during preview

berhalak:
When an inner widget was reporting some mandatory columns to map, the builder was causing infinite loop. More in comment, but here is summary.
- Builder was sending ready message to Grist, without any mapped columns
- Grist was rendering the widget, sending JS/HTML back
- Builder was creating inner widget, which called ready method and reported some mandatory columns to map
- Grist unloads builder, as columns where not mapped
- User mapped a column
- Grist loads builder
- Builder reports back to Grist that no columns to map (and no mandatory columns anymore). 
- Grist unloads builder once again (as mandatory option has changed, and Grist is sensitive here).
- Grist sees no mandatory columns, so shows builder once again.
- Builder, reports back no mandatory columns, creates inner widget, this inner widget reports mandatory columns
- And it loops

Here the basic idea was not to call the `configure` endpoint from the builder, instead builder is injecting its configuration in the `configure` call from the inner widget, that way Grist sees only one configure call, and columns mapping stays consistent during reloads.

Not related:
- Some dead code is removed
- api_deps.js is loaded with monaco
- moncaco scripts are loaded in parallel 

Here is a document with this builder and markdown widget's content inside (with mandatory column):
https://public.getgrist.com/d1t1x4BvxuMK/JsFiddle

### DIFF
--- a/custom-widget-builder/api.js
+++ b/custom-widget-builder/api.js
@@ -385,7 +385,7 @@ window.addEventListener('message', e => {
     ) {
       e.data.args ??= [{}];
       e.data.args[0].hasCustomOptions = true;
-      // Here is the trick part. Inner widget is calling ready method and configure method (as
+      // Here is the tricky part. Inner widget is calling ready method and configure method (as
       // part of grist.ready() call). Since we don't call configure method at all, we just append
       // the `hasCustomOptions` to the configure method, and send it back to Grist. This way Grist
       // will receive same configuration of mapping columns each time, which will avoid the infinite
--- a/custom-widget-builder/api.js
+++ b/custom-widget-builder/api.js
@@ -180,42 +180,14 @@ function purge(element) {
   }
 }
 
-let lastListener;
+let widgetWindow = null;
 function createFrame() {
   // remove all data from page_widget
   purge(page_widget);
   wFrame = document.createElement('iframe');
   page_widget.appendChild(wFrame);
-  const widgetWindow = wFrame.contentWindow;
-  // Rewire messages between this widget, and the preview
-  if (lastListener) window.removeEventListener('message', lastListener);
-  lastListener = e => {
-    if (e.source === widgetWindow) {
-      // Hijicack configure message to inform Grist that we have custom configuration.
-      // data will have { iface: "CustomSectionAPI", meth: "configure", args: [{}] }
-      if (
-        e.data?.iface === 'CustomSectionAPI' &&
-        e.data?.meth === 'configure'
-      ) {
-        e.data.args ??= [{}];
-        e.data.args[0].hasCustomOptions = true;
-      }
-      window.parent.postMessage(e.data, '*');
-    } else if (e.source === window.parent) {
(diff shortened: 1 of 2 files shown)


custom-widget-builder/api.js
```diff
@@ -369,16 +319,20 @@ onOptions(async options => {
...
     _html: htmlModel.getValue(),
   };
-  grist.setOptions(options);
+
   state('installed');
 
-  // Copy file contents into memory.
-  currentJs(options?._js);
-  currentHtml(options?._html);
+  // Compare options with lastOptions, and if they are the same, don't call setOptions.
```
berhalak, line 328: If nothing has changed, there is no need to save it.

custom-widget-builder/api.js
```diff
@@ -1,35 +1,21 @@
-/** Helper for keeping some data and watching for changes */
 function memory(name) {
   let value = undefined;
-  let listeners = [];
```
berhalak, line 4: Not used anywhere (anymore).

custom-widget-builder/api.js
```diff
@@ -1,35 +1,21 @@
...
-
   return obj;
 }
 
 // Global state, to keep track of the editor state, and file content.
 const currentJs = memory('js');
 const currentHtml = memory('html');
 const state = memory('state'); // null, 'installed', 'editor'
-
-const COLORS = {
```
berhalak, line 29: Dead code

custom-widget-builder/api.js
```diff
@@ -395,10 +349,59 @@ function bntReset_onClick() {
...
+  if (e.source === widgetWindow) {
+    // Hijack configure message to inform Grist that we have custom configuration.
+    // data will have { iface: "CustomSectionAPI", meth: "configure", args: [{}] }
+    if (
+      e.data?.iface === 'CustomSectionAPI' &&
+      e.data?.meth === 'configure'
+    ) {
+      e.data.args ??= [{}];
+      e.data.args[0].hasCustomOptions = true;
+      // Here is the trick part. Inner widget is calling ready method and configure method (as
```
georgegevoian, line 388:
```suggestion
      // Here is the tricky part. Inner widget is calling ready method and configure method (as
```

custom-widget-builder/api.js
```diff
@@ -395,10 +349,59 @@ function bntReset_onClick() {
...
+      // part of grist.ready() call). Since we don't call configure method at all, we just append
+      // the `hasCustomOptions` to the configure method, and send it back to Grist. This way Grist
+      // will receive same configuration of mapping columns each time, which will avoid the infinite
+      // loop problem described above.
     }
-  },
+    window.parent.postMessage(e.data, '*');
+  } else if (e.source === window.parent) {
+    // If user clicks `Open confirguration` button, we will switch to the editor.
```
georgegevoian, line 396:
```suggestion
    // If user clicks `Open configuration` button, we will switch to the editor.
```

georgegevoian (approved):

PR #147, opened 2024-12-30
invoice: add a PAID stamp

jordigh:
If the invoice has a Paid toggle, this will show a little "PAID" stamp on the invoice once that's checked.


Spoffy (changes requested):
Looks like "Paid" is a new column, so I think we need to add it to the script's list of "valid" columns:
image

As currently it shows up as ignored!

We could also add it to `exampleData.js`?

Also a really nice looking stamp  :slightly_smiling_face:

jordigh:
Thanks! I made those changes.

PR #149, opened 2025-01-07
Adding choice list in timeline widget

berhalak:
Adding support for ChoiceList column.

Not related:
- removing some dead code
- displaying labels in headers instead of colIds


berhalak:
Since this is not yet published widget, I'm merging it without a formal review.

PR #151, opened 2025-01-12
Fix geocoder to new 3.x API

SevereOverfl0w:
As index.html is using the latest version of leaflet-control-geocoder and the API changed in 3.0.0 with this example:

```diff
Migrate from callbacks to Promise

-geocoder.geocode(query, callback);
+geocoder.geocode(query).then(callback);

-geocoder.reverse(latlng, scale, callback);
+geocoder.reverse(latlng, scale).then(callback);

Migrate from callbacks to async-await

-geocoder.geocode(query, callback);
+const results = await geocoder.geocode(query);

-geocoder.reverse(latlng, scale, callback);
+const results = await geocoder.reverse(latlng, scale);
```

The usage in the map widget also needs updating to correspond.

### DIFF
--- a/map/page.js
+++ b/map/page.js
@@ -103,18 +103,14 @@ if (URLSearchParams && location.search && geocoder) {
 }
 
 async function geocode(address) {
-  return new Promise((resolve, reject) => {
-    try {
-      geocoder.geocode(address, (v) => {
-        v = v[0];
-        if (v) { v = v.center; }
-        resolve(v);
-      });
-    } catch (e) {
-      console.log("Problem:", e);
-      reject(e);
-    }
-  });
+  const results = await geocoder.geocode(address);
+  let v = results[0];
+
+  if (v) {
+    v = v.center;
+  }
+
+  return v;
 }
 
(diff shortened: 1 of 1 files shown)


map/page.js
```diff
@@ -103,18 +103,14 @@ if (URLSearchParams && location.search && geocoder) {
...
-        v = v[0];
-        if (v) { v = v.center; }
-        resolve(v);
-      });
-    } catch (e) {
-      console.log("Problem:", e);
-      reject(e);
-    }
-  });
+  const results = await geocoder.geocode(address);
```
jordigh, line 106: This looks reasonable, but it seems like the original `try`-`catch` should be kept. Why is it no longer needed?

SevereOverfl0w: I dropped it because it should just bubble out now due to using await, rather than needing explicit handling to conform it to the promise. But I can reinstate it with the logging out of caution, no problem.

paulfitz (approved):
Thanks @SevereOverfl0w ! Dropping the try/catch seems fine.

PR #154, opened 2025-03-25
tests: restore gristWebDriverUtils from grist-core

jordigh:
As the header of that file says, it should have never been independently changed. Lots of changes happened to it regardless since the file was copied.

As a first step to fixing the tests, this restores the original copy from grist-core and moves all of the additions to a supplemental subclass.

### DIFF
--- a/test/getGrist.ts
+++ b/test/getGrist.ts
@@ -4,10 +4,8 @@ import fs from 'fs';
 import { driver, enableDebugCapture } from 'mocha-webdriver';
 import fetch from 'node-fetch';
 
-import {GristWebDriverUtils} from 'test/gristWebDriverUtils';
-
-
-
+import {Key} from 'mocha-webdriver';
+import {GristWebDriverUtils} from "test/gristWebDriverUtils";
 
 /**
  * Set up mocha hooks for starting and stopping Grist. Return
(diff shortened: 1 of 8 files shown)


paulfitz (dismissed):
Thanks @jordigh ! Very tiny lint. Also upstream gristWebDriverUtils has updated very slightly in the meantime, at least there's a small difference.

test/getGrist.ts
```diff
@@ -285,4 +286,244 @@ export class GristUtils extends GristWebDriverUtils {
...
+  }
+
+
+  /**
+   * Returns a visible GridView cell. Options may be given as arguments directly, or as an object.
+   * - col: column name, or 0-based column index
+   * - rowNum: 1-based row numbers, as visible in the row headers on the left of the grid.
+   * - section: optional name of the section to use; will use active section if omitted.
+   */
+  public getCell(col: number | string, rowNum: number, section?: string): WebElementPromise;
```
paulfitz, line 423: On second thought, some of these methods are very confusing! There's nothing grist-widget-y about them, how come regular methods written in grist-core don't cover them?

Hang on, is someone copying and pasting from grist-core to grist-widget? Nooooo...

Ok, let me undo my review...

paulfitz: Ok, @jordigh, here's the situation. It looks like somewhere along the line someone has been copying code from `test/nbrowser/gristUtils.ts` in `grist-core` into `grist-widget` and reviewers have been letting this happen. NO. We can't afford to have two copies of this kind of code. It will get updated in `grist-core`, but it won't get updated in `grist-widget` and tests will over time stop working, they couldn't possibly work as the test harness code and the app get out of sync.

What needs to happen, when people need test harness code from `grist-core`, is that first it gets factored out. The way we have to do that (lame! but workable if people review for it!) is to first move the code to `gristWebDriverUtils.ts` and copy that across. That way, if tests fail, at least the shared harness code is updatable mechanically. Jarek and George or Dmitry or me or anyone are never going to think to update copy/pasted stuff, or even if we thought about it, actually know how to do it outside of grist-core with all its rigorous tests.

For for this task of getting grist-widget back to sanity, you need to evaluate every method you are sticking here, see if it got copied from grist-core (likely gristUtils.ts) there, and refactor grist-core if so to move it to gristWebDriverUtils. Does that make sense to you?

paulfitz (commented):
Hi @jordigh, there's a remaining problem with copy/pasted code that could well be stale.

paulfitz (approved):
Great, thanks @jordigh ! Given this, I looked at a random test failure (the first chart one) and it was an easy fix (adding a `await driver.findContentWait('button', 'Hide', 200).click();` for a button that looks important to click) so I'm optimistic for follow-up.

PR #155, opened 2025-03-27
[Map] Fix geocode bug

headphoner:
Currently in the map widget, failure to geocode an address for one row causes all rows below it to no longer geocode. This proposed change resolves the underlying unhandled null reference.

### Old behavior
https://github.com/user-attachments/assets/480d69e2-9225-4528-8f1e-11353c6c8a44

### New behavior
https://github.com/user-attachments/assets/0f0ff9ba-315f-44e8-9069-c00af39557bc


map/page.js
```diff
@@ -137,20 +137,25 @@ async function scan(tableId, records, mappings) {
...
     if (address && !record[Longitude]) {
       // Find coordinates.
       const result = await geocode(address);
       // Update them, and update cache (if the field was mapped)
       await grist.docApi.applyUserActions([ ['UpdateRecord', tableId, record.id, {
-        [mappings[Longitude]]: result.lng,
-        [mappings[Latitude]]: result.lat,
+        [mappings[Longitude]]: result?.lng,
+        [mappings[Latitude]]: result?.lat,
         ...(GeocodedAddress in mappings) ? {[mappings[GeocodedAddress]]: address} : undefined
```
berhalak, line 159: This is not related, but I have trouble running this geocoding feature due to this line. For me `mappings[GeocodedAddress]` is null and this line fails. Can you check that the column is truly mapped?
```suggestion
        ...(GeocodedAddress in mappings && mappings[GeocodedAddress]) ? {[mappings[GeocodedAddress]]: address} : undefined
```

map/page.js
```diff
@@ -137,20 +137,25 @@ async function scan(tableId, records, mappings) {
...
     }
     // If address is not empty, and coordinates are empty (or were cleared by cache)
     if (address && !record[Longitude]) {
       // Find coordinates.
       const result = await geocode(address);
       // Update them, and update cache (if the field was mapped)
       await grist.docApi.applyUserActions([ ['UpdateRecord', tableId, record.id, {
-        [mappings[Longitude]]: result.lng,
-        [mappings[Latitude]]: result.lat,
+        [mappings[Longitude]]: result?.lng,
```
berhalak, line 157:
```suggestion
        [mappings[Longitude]]: result?.lng ?? null,
```

map/page.js
```diff
@@ -137,20 +137,25 @@ async function scan(tableId, records, mappings) {
...
+        [mappings[Latitude]]: result?.lat,
```
berhalak, line 158:
```suggestion
        [mappings[Latitude]]: result?.lat ?? null,
```

berhalak (changes requested):
Thanks for the fix @headphoner! Tiny comments for null/undefined handling.

headphoner:
Thanks for the comments @berhalak . I've merged them into the PR. Please review and let me know what you think.

berhalak (approved):
Thanks!

PR #161, opened 2025-04-24
Improve calendar events display

mperreir:
Make events text wrap in their container so that the whole text can more easily be seen


vviers (approved):
LGTM

PR #164, opened 2025-06-04
Updating tests

berhalak:
Updating bunch of tests just to make them work.

Other changes:
- Fixing release workflow, it wasn't installing node dependency (now needed for publishing)
- One github action was replaced with a fork, related issue: https://github.com/ScribeMD/docker-cache/issues/837.

### DIFF
--- a/test/getGrist.ts
+++ b/test/getGrist.ts
@@ -78,6 +78,12 @@ export class GristTestServer {
       );
     }
     const pwd = process.cwd();
+
+    // Make sure rewriteUrl.js is available in the buildtools directory.
+    if (!fs.existsSync(`${pwd}/buildtools/rewriteUrl.js`)) {
+      throw new Error(`Expected buildtools/rewriteUrl.js to exist at ${pwd}/buildtools/rewriteUrl.js`);
+    }
+
     this._assetServer = spawn('live-server', [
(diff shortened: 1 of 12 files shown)


test/getGrist.ts
```diff
@@ -198,13 +204,24 @@ export class GristUtils extends GristWebDriverUtils {
     await this.waitForServer();
   }
 
-  public async clickWidgetPane() {
+  // This method doesn't work in single user mode, need to fix upstream.
+  // For now it is  just overridden with a simpler version.
```
jordigh, line 208: Can you open a task to do this so it's not completely forgotten? We need to keep the test harness in synch as much as possible between core and widgets.

test/markdown.ts
```diff
@@ -35,7 +35,7 @@ describe('markdown', function() {
       await grist.driver.sendKeys(Key.END, Key.ENTER, Key.ENTER,
         ' - Phone Number: (123) 456-7890');
     });
-    await grist.openAccountMenu();
+    await grist.driver.findWait('.test-dm-account', 2000).click();
```
jordigh, line 38: Is this change still necessary even after overriding `openAccountMenu`?

test/getGrist.ts
berhalak: It is overridden, so this is still in sync.

PR #175, opened 2025-10-16
Sanitizing preview in markdown widget. Fixing map widget selections when new record is clicked.

berhalak:
- Make Markdown widget sanitize data. Currently it allows scripts to run from cells (e.g. via content like ``).
- Fixing map selection when a new row is clicked.

### DIFF
--- a/map/page.js
+++ b/map/page.js
@@ -194,7 +194,7 @@ function getInfo(rec) {
 }
 
 // Function to clear last added markers. Used to clear the map when new record is selected.
-let clearMakers = () => {};
+let clearMarkers = () => {};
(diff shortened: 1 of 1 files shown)


map/page.js
```diff
@@ -403,14 +413,21 @@ grist.onRecords((data, mappings) => {
 });
 
 grist.onNewRecord(() => {
-  clearMakers();
-  clearMakers = () => {};
+  if (mode === 'single') {
+    clearMakers();
```
dsagal: Maybe a good chance to rename it to "clearMarkers"?

PR #176, opened 2025-10-17
Fix <style> tags in Markdown widget.

dsagal:
- The issue was that DOMPurify default has the poor legacy default of `{WHOLE_DOCUMENT: false, FORCE_BODY: false}`, whose effect is to hoist the <style> tag at the beginning of content into document \<head> and then discard this \<head>. The effect is that the first <style> element is lost. It's not even the result of sanitization.

  Setting `FORCE_BODY: true` fixes it.

- Also upgrade DOMPurify CDN version to the same one used in the current version of Grist app itself.


paulfitz (approved):
Thanks for figuring this out @dsagal !

PR #181, opened 2025-11-20
Page builder: Create pages or catalogs from Grist data.

dsagal:


dsagal:
This will be approached a bit differently.

PR #198, opened 2026-03-11
fix(calendar): open on today's date by default on initial load

lmzr:
### Problem

When the Grist calendar widget was first loaded, it would navigate to
the date of the **first selected record** rather than showing today's
date. If the record was months in the past or future, the user had to
manually navigate back to the current date.

### Root cause

`gristSelectedRecordChanged()` is called immediately at startup with
whatever record Grist considers "selected". The existing code called
`calendarHandler.selectRecord()` unconditionally, which scrolled the
view to that record's date.

### Fix

Added a one-shot `isInitialLoad` flag (initialized to `true`).

On the **first** call to `gristSelectedRecordChanged()`:
- The flag is cleared (`isInitialLoad = false`)
- `calendarHandler.calendarToday()` is called to land on today
- The function returns early (no record navigation)

All subsequent calls work as before: the view follows the selected record.

### Files changed

- `calendar/page.js` — `isInitialLoad` flag + guard in `gristSelectedRecordChanged()`

### Testing

1. Open the calendar widget with records spread across multiple months
2. Confirm the widget opens on **today's date**, not on the first record
3. Click a record in the Grist table → confirm the calendar navigates to that record
4. Reload the page → confirm the widget opens on today again

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -20,7 +20,6 @@ window.gristCalendar = {
 let TZDate = null;
 
 let isInitialLoad = true;
-let startOnToday = true;
 
 function getLanguage() {
   if (this._lang) {
@@ -559,20 +558,12 @@ async function translatePage() {
 function gristSelectedRecordChanged(record, mappings) {
   const mappedRecord = grist.mapColumnNames(record, mappings);
   if (mappedRecord && calendarHandler) {
-    if (isInitialLoad && startOnToday) {
+    if (isInitialLoad) {
       isInitialLoad = false;
       calendarHandler.calendarToday();
       return;
     }
-    isInitialLoad = false;
     calendarHandler.selectRecord(mappedRecord);
-  }
-}
-
-async function startOnTodayChanged(checkbox) {
-  startOnToday = checkbox.checked;
-  if (!isReadOnly) {
-    await grist.setOption('startOnToday', startOnToday);
   }
 }
 
@@ -592,9 +583,7 @@ function onGristSettingsChanged(options, settings) {
   const view = options?.calendarViewPerspective ?? 'week';
   changeCalendarView(view);
   colTypesFetcher.setAccessLevel(settings.accessLevel);
-  startOnToday = options?.startOnToday ?? true;
-  const checkbox = document.getElementById('calendar-start-on-today');
-  if (checkbox) { checkbox.checked = startOnToday; }
+
 };
 
 function changeCalendarView(view) {
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -19,6 +19,9 @@ window.gristCalendar = {
 
 let TZDate = null;
 
+let isInitialLoad = true;
+let startOnToday = true;
+
 function getLanguage() {
   if (this._lang) {
     return this._lang;
@@ -556,7 +559,20 @@ async function translatePage() {
 function gristSelectedRecordChanged(record, mappings) {
   const mappedRecord = grist.mapColumnNames(record, mappings);
   if (mappedRecord && calendarHandler) {
+    if (isInitialLoad && startOnToday) {
+      isInitialLoad = false;
+      calendarHandler.calendarToday();
+      return;
+    }
+    isInitialLoad = false;
     calendarHandler.selectRecord(mappedRecord);
+  }
+}
+
+async function startOnTodayChanged(checkbox) {
+  startOnToday = checkbox.checked;
+  if (!isReadOnly) {
+    await grist.setOption('startOnToday', startOnToday);
   }
 }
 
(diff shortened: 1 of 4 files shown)


berhalak (changes requested):
Hi @lmzr, once https://github.com/gristlabs/grist-core/pull/2259 lands, this PR can use the information about linking state, and in case there is no linking, we can safely set today's date on start.

fflorent:
Hi @lmzr,

Do you need help on this PR?

lmzr:
Hi @berhalak @paulfitz @fflorent — sorry for the delay, and thanks for your patience.

Now that grist-core#2259 has landed (linking info is available since Grist v1.7.13), I've reworked this PR to use it instead of blindly setting today's date.

The widget now reads `settings.linking` from `onOptions` and decides per cursor event whether to move the viewport:
- **no incoming link** (`asTarget === null`) → open on today and only highlight the cursor record, without ever moving the view;
- **driven by a link** (`asTarget !== null`), or an older Grist that doesn't report linking → follow the selected record, exactly as before.

This addresses the original concern: linked sections keep behaving correctly, and we only force today when it's safe to do so. It also fixes a subtler issue I found while testing — previously the view jumped to the first record's date whenever the cursor was re-emitted (e.g. after saving the perspective option), which is now avoided in the non-linked case.

The branch is rebased on current master and the deploy preview is green.

calendar/page.js
```diff
@@ -302,18 +324,26 @@ class CalendarHandler {
...
     }
 
     if (this._selectedRecordId) {
       this._clearHighlightEvent(this._selectedRecordId);
     }
+    this._selectedRecordId = record.id;
+
+    // When the calendar isn't driven by a link, keep the selection highlighted
+    // but never move the viewport to the record's date (see gristSelectedRecordChanged).
+    if (!navigate) {
```
fflorent, line 339: Any reason for introducing `navigate` here? As its value, according to the comment above, seems bound to `isDrivenByLink()`.

calendar/page.js
```diff
@@ -556,7 +586,36 @@ async function translatePage() {
...
+  if (!linkingResolved || !pendingInitialRecord) { return; }
+  isInitialLoad = false;
+  const record = pendingInitialRecord;
+  pendingInitialRecord = null;
+  if (isDrivenByLink()) {
+    calendarHandler.selectRecord(record, true);
+  } else {
+    calendarHandler.calendarToday();
+    calendarHandler.selectRecord(record, false);
   }
```
fflorent: If keeping `navigate` (see my comment above), boolean flag lacks of semantic. I suggest to either turn the argument to an option (`navigate: boolean` => `options: { navigate: boolean }`) or change the code above to something like this:


```suggestion
  if (!isDrivenByLink()) {
    calendarHandler.calendarToday();
  }
  
  calendarHandler.selectRecord(record, isDrivenByLink());
  
```

fflorent (approved):
Minor comments on my side, otherwise looks good to me, thanks!

calendar/page.js
lmzr: You're right that it always equals `isDrivenByLink()` at both call sites. I kept the parameter (it defaults to `true`, so any future caller keeps the old behaviour) but removed the opaque boolean literals — `applyInitialView` now passes `isDrivenByLink()` through. Addressed in be06f61.

calendar/page.js
lmzr: Applied your suggestion — kept `navigate`, but `applyInitialView` now derives it once from `isDrivenByLink()` and only jumps to today when there's no link, so the duplicated branch and the bare `true`/`false` are gone. be06f61. Thanks!

ornanovitch:
what is the situation here?

berhalak:
> what is the situation here?

Calendar is now part of the product as a native widget, so I needed to freeze for a moment code changes to this widget. I'm doing the review right now, sorry for the delay.

calendar/page.js
```diff
@@ -19,6 +19,28 @@ window.gristCalendar = {
...
+// the view: the calendar opens on today and stays where the user navigates,
+// otherwise it would jump to the first record's date on load and on every
+// re-emitted cursor event (e.g. after saving the perspective option).
+// Grist reports linking via `settings.linking` in onOptions (grist-core#2259,
+// available since Grist v1.7.13). Since onOptions and onRecord can fire in either
+// order, we defer the initial decision until both are known.
+let isInitialLoad = true;
+let linkingResolved = false;   // true once onOptions has fired at least once
+let linkingInfo;               // settings.linking: {asTarget, asSource} | undefined on Grist < 1.7.13
+let pendingInitialRecord = null;
```
berhalak, line 38: Can you add some comments about this variable?

berhalak (approved):
Looks good and works ok, just this one nitpick with the comment update.

calendar/page.js
lmzr: Done in cf7f09a: added comments describing each of the initial-view state variables, in particular `pendingInitialRecord` (the first record buffered until linking info arrives from onOptions). Thanks for the review!

PR #199, opened 2026-03-11
feat(calendar): detect first day of week from locale

lmzr:
### Problem

The Grist calendar widget always displayed weeks starting on **Sunday**,
regardless of the user's locale. In most European and many other locales,
weeks should start on **Monday**.

### Fix

Added a `getFirstDayOfWeek()` function that reads the first weekday from
the browser's `Intl.Locale` API:

- Reads locale from the Grist `culture` URL parameter first, then falls
  back to `navigator.language` and finally to the widget's active language
- Uses `locale.getWeekInfo()` (Chrome 99+) or `locale.weekInfo` (Firefox
  126+, Safari 17+) to retrieve the locale's first weekday
- Converts from Intl convention (1=Mon…7=Sun) to TUI Calendar convention
  (0=Sun, 1=Mon…6=Sat)
- Falls back to `0` (Sunday) if the API is not supported

`startDayOfWeek` is set on both `week` and `month` calendar options so all
views are consistent.

A follow-up commit (`fix: use Grist culture param for first day of week
detection`) ensures the Grist `culture` URL parameter is prioritised, so
the calendar respects the language/region configured in the Grist document
settings rather than the browser OS locale alone.

### Files changed

- `calendar/page.js` — `getFirstDayOfWeek()` function + `startDayOfWeek`
  in `week` and `month` options

### Testing

1. Open the widget with a French (`fr-FR`) Grist culture → weeks start on Monday
2. Open with `en-US` culture → weeks start on Sunday
3. Verify behaviour is the same in week view and month view
4. In a browser that does not support `Intl.Locale.getWeekInfo` → fallback
   to Sunday, no JS error

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -53,6 +53,21 @@ function getMonthName() {
   return calendarHandler.calendar.getDate().toDate().toLocaleString(getLanguage(), {month: 'long', year: 'numeric'})
 }
 
+function getFirstDayOfWeek() {
+  try {
+    const locale = new Intl.Locale(getLanguage());
+    // Chrome 99+ uses getWeekInfo(), Firefox 126+ and Safari 17+ use weekInfo property
+    const weekInfo = locale.getWeekInfo?.() ?? locale.weekInfo;
+    if (weekInfo?.firstDay !== undefined) {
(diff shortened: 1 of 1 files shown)


paulfitz (approved):
thanks @lmzr !

PR #200, opened 2026-03-11
feat(calendar): detect 12h/24h time format from locale

lmzr:
### Problem

The Grist calendar widget always displayed time slots in the day/week view
using the **24-hour format** (e.g. `14:00`), even for locales like `en-US`
where 12-hour format (e.g. `2 PM`) is standard.

### Fix

Added a `getHour12()` function that detects the preferred time format for
the active locale:

- Reads locale from the Grist `culture` URL parameter, then falls back to
  `navigator.language` and the widget's active language
- Uses `Intl.DateTimeFormat.resolvedOptions().hour12` to determine whether
  the locale prefers 12-hour (`true`) or 24-hour (`false`) format
- Falls back to `false` (24h) on error or if the API is unsupported

The result is applied via the `timegridDisplayPrimaryTime` template
callback in TUI Calendar options, formatting each time label using
`Date.toLocaleTimeString()` with the detected `hour12` value.

### Files changed

- `calendar/page.js` — `getHour12()` function + `timegridDisplayPrimaryTime`
  callback in calendar options

### Testing

1. Open the widget with `en-US` culture → time labels show `1 AM`, `2 PM`, etc.
2. Open with `fr-FR` culture → time labels show `01:00`, `14:00`, etc.
3. Verify no regression on event creation/editing popup times
4. In a browser that does not support `Intl.DateTimeFormat` → fallback to
   24h, no JS error

### DIFF
--- a/calendar/index.html
+++ b/calendar/index.html
@@ -53,15 +53,22 @@
 <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
 <script src="https://uicdn.toast.com/tui.time-picker/latest/tui-time-picker.js"></script>
 <script>
-  // HACK: Override the default time picker format to be 24-hour. There is no other way to do this,
+  // HACK: Override the default time picker format based on locale. There is no other way to do this,
   // calendar doesn't offer a way to pass this down to the time picker.
   // This also can't be done in the page.js as the next script will grab the reference immediately.
   (function() {
+    var useHour12 = false;
+    try {
+      var culture = new URLSearchParams(window.location.search).get('culture')
+        || navigator.language
+        || 'en';
+      useHour12 = new Intl.DateTimeFormat(culture, {hour: 'numeric'}).resolvedOptions().hour12;
+    } catch(e) {}
     tui.TimePicker = class extends tui.TimePicker {
       constructor(container, options) {
         super(container, {
           ...options,
-          format: 'HH:mm'
+          format: useHour12 ? 'hh:mm A' : 'HH:mm'
         });
       }
     };
--- a/calendar/index.html
+++ b/calendar/index.html
@@ -53,9 +53,14 @@
 <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
 <script src="https://uicdn.toast.com/tui.time-picker/latest/tui-time-picker.js"></script>
 <script>
-  // HACK: Override the default time picker format based on locale. There is no other way to do this,
-  // calendar doesn't offer a way to pass this down to the time picker.
-  // This also can't be done in the page.js as the next script will grab the reference immediately.
+  // HACK: Override the time picker format based on locale.
+  // This targets the TimePicker inside the event create/edit popup, which is a
+  // separate widget from the time-grid axis handled by timegridDisplayPrimaryTime
+  // in page.js: that template only formats the hour labels of the week/day grid,
+  // not the popup input. The TimePicker defaults to a 12-hour format, so without
+  // this override 24-hour locales would still get an AM/PM input. There is no API
+  // to pass the format down to it, and it can't be done in page.js as the next
+  // script grabs the tui.TimePicker reference immediately.
   (function() {
     var useHour12 = false;
     try {
--- a/calendar/index.html
+++ b/calendar/index.html
@@ -51,6 +51,7 @@
 </div>
 </body>
 <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
+<script src="hour-format.js"></script>
 <script src="https://uicdn.toast.com/tui.time-picker/latest/tui-time-picker.js"></script>
 <script>
   // HACK: Override the time picker format based on locale.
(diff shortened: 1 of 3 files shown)


calendar/index.html
```diff
@@ -53,15 +53,22 @@
 <script src="https://docs.getgrist.com/grist-plugin-api.js"></script>
 <script src="https://uicdn.toast.com/tui.time-picker/latest/tui-time-picker.js"></script>
 <script>
-  // HACK: Override the default time picker format to be 24-hour. There is no other way to do this,
+  // HACK: Override the default time picker format based on locale. There is no other way to do this,
```
berhalak, line 56: Have you tested a solution where we just remove this hack? It seems to work ok for me. I'm not sure why this hack is even here now, as you clearly are passing this down to the calendar in this timegridDisplayPrimayTime method.

calendar/page.js
```diff
@@ -53,6 +53,16 @@ function getMonthName() {
   return calendarHandler.calendar.getDate().toDate().toLocaleString(getLanguage(), {month: 'long', year: 'numeric'})
 }
 
+function getHour12() {
+  try {
+    const locale = urlParams.get('culture') ?? navigator.language ?? getLanguage();
+    return new Intl.DateTimeFormat(locale, {hour: 'numeric'}).resolvedOptions().hour12;
+  } catch (e) {
+    // Intl.DateTimeFormat not supported or invalid locale
```
berhalak, line 61: This should show something in the console at least.

calendar/page.js
```diff
@@ -53,6 +53,16 @@ function getMonthName() {
   return calendarHandler.calendar.getDate().toDate().toLocaleString(getLanguage(), {month: 'long', year: 'numeric'})
 }
 
+function getHour12() {
+  try {
+    const locale = urlParams.get('culture') ?? navigator.language ?? getLanguage();
```
berhalak, line 58: Can you make this function pure?

calendar/page.js
```diff
@@ -184,8 +194,14 @@ class CalendarHandler {
         },
         popupIsAllday() {
           return t('All Day')
-        }
-
+        },
+        timegridDisplayPrimaryTime({ time }) {
```
berhalak, line 198: Can you add some comment what this function does and what is the intent here, and how lanugage is translated to locale ?

berhalak (changes requested):

fflorent:
Hi @lmzr,

Do you need help to make this PR land?

calendar/index.html
lmzr: Good question — I did look at removing it. The catch is that this override targets the **TimePicker in the create/edit event popup**, which is a different widget from the time-grid axis. `timegridDisplayPrimaryTime` only formats the hour labels on the left of the week/day grid; it doesn't reach the popup input.

The TimePicker defaults to a 12-hour format, so dropping the override would give 24-hour locales (e.g. `fr-FR`) an AM/PM input in the popup. It probably looked fine on your side because a 12-hour locale gets the format it wants either way.

So I kept the hack, but it's now locale-driven instead of hardcoded to 24h. I've also reworded the comment in `index.html` to spell out why it's still needed and how it differs from the grid template.

calendar/index.html
```diff
@@ -53,15 +53,27 @@
...
+  // to pass the format down to it, and it can't be done in page.js as the next
+  // script grabs the tui.TimePicker reference immediately.
   (function() {
+    var useHour12 = false;
+    try {
+      var culture = new URLSearchParams(window.location.search).get('culture')
+        || navigator.language
+        || 'en';
+      useHour12 = new Intl.DateTimeFormat(culture, {hour: 'numeric'}).resolvedOptions().hour12;
+    } catch(e) {}
```
fflorent: Sounds like the same additions to page.js?

If so, do you think it could be feasible to factorize a bit the code?

calendar/page.js
```diff
@@ -68,6 +68,25 @@ function getFirstDayOfWeek() {
...
 }
 
+// Resolve the BCP 47 locale tag to use for all locale-aware formatting.
+// Grist passes the document language as the 'culture' URL param (e.g. "fr-FR");
+// we fall back to the browser language, then to the i18next-detected language.
+function getLocale() {
+  return urlParams.get('culture') ?? navigator.language ?? getLanguage();
+}
+
+// Pure: given a BCP 47 locale tag, returns true when that locale's convention
```
fflorent: Maybe commenting that the function is pure is a bit overkill :)

fflorent (approved):
LGTM, few comments on my side, I let Grist Labs review on their side.

calendar/index.html
lmzr: Done for the part that was genuinely duplicated — the 12h/24h detection is now `getHour12()` in a small shared `hour-format.js`, loaded before both the inline TimePicker patch and page.js and called from each. I kept the locale resolution inline in the two places since they legitimately differ (page.js's `getLocale` has an i18next fallback that isn't available yet when the TimePicker patch runs). c0772d1.

calendar/page.js
lmzr: Agreed — dropped it (it wasn't even accurate: the function `console.warn`s on failure). c0772d1.

PR #201, opened 2026-03-11
feat(calendar): hide time grid when start/end columns are pure Date type

lmzr:
### Problem

When the Grist calendar widget is configured with `Date` columns (without
time component) for Start Date and/or End Date, the day/week view still
displayed the full **hourly time grid**, wasting most of the visible area
for a tiny "All Day" strip at the top. Additionally, if more than 3
all-day events existed, TUI Calendar capped the display at 3 rows and
showed "+N more" overflow buttons without auto-expanding.

### Fix

**Pure-date detection** (`calendar/page.js` — `updateCalendar()`):
- After fetching column types, compute `isPureDate = startType === 'Date'
  && (!endType || endType === 'Date')`
- Toggle the CSS class `pure-date-mode` on the `#calendar` element

**CSS rules** (`calendar/screen.css`):
- `.pure-date-mode .toastui-calendar-time` → `display: none` hides the
  hourly grid entirely
- `.pure-date-mode .toastui-calendar-allday` → `min-height: calc(100% - 44px)`
  makes the all-day panel fill the freed space
- `.pure-date-mode .toastui-calendar-weekday-exceed-in-week` → `display: none`
  hides the "+N more" buttons (auto-expand is handled by JS)

**MutationObserver** (`setupAlldayAutoExpand()`):
- In pure-date mode, a `MutationObserver` watches for overflow buttons
  injected by TUI and auto-clicks them via `requestAnimationFrame` so all
  events are always visible without user interaction
- The observer is properly disconnected when leaving pure-date mode

**Guard in scroll handler** (`calendar/page.js` line 307):
- Added an early return when `.toastui-calendar-time` container is absent
  (i.e. in pure-date mode), preventing a null-reference error in the
  existing scroll-into-view logic

### Files changed

- `calendar/page.js` — `isPureDate` detection, `setupAlldayAutoExpand()`,
  null-guard in scroll handler
- `calendar/screen.css` — 3 CSS rules for `pure-date-mode`

### Behaviour matrix

| startType | endType     | isPureDate | Result                          |
|-----------|-------------|------------|---------------------------------|
| `Date`    | `Date`      | true       | Time grid hidden, all-day fills |
| `Date`    | (unmapped)  | true       | Same                            |
| `Date`    | `DateTime`  | false      | Normal time grid                |
| `DateTime`| `Date`      | false      | Normal time grid                |
| `DateTime`| `DateTime`  | false      | Unchanged behaviour             |

Dynamic reconfiguration is automatic: `updateCalendar()` is re-called on
every mapping change, so switching column types between `Date` and
`DateTime` updates the view immediately without a page reload.

### Testing

1. Configure the widget with `Date` columns (Start + End) → week view shows
   only the all-day strip, no hourly grid
2. With many events on the same day → all events visible, no "+N more" button
3. Create an event via double-click → popup has no time fields
4. Switch columns to `DateTime` → hourly grid reappears immediately
5. Month view → unaffected (no time grid to hide)
6. Existing tests in `test/calendar.ts` → all pass

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -255,6 +255,9 @@ class CalendarHandler {
     // Deletion happens via the event-edit form.
     this.calendar.on('beforeDeleteEvent', (eventInfo) => deleteEvent(eventInfo));
 
+    // Keep the pure-date all-day panel expanded after every render (see method).
+    this.calendar.on('afterRenderEvent', () => this._expandPureDateAllDayPanel());
+
     container.addEventListener('mousedown', () => {
       focusWidget();
       // Clear existing selection; this follows the suggested workaround in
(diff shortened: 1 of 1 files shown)


manuhabitela:
Hey there :) thanks for the PR.

I have to say when first reading this I felt the whole MutationObserver thing was a bit… much. But at the same time it seems difficult to have the behavior you want without doing it like this 😅

berhalak:
> Hey there :) thanks for the PR.
> 
> I have to say when first reading this I felt the whole MutationObserver thing was a bit… much. But at the same time it seems difficult to have the behavior you want without doing it like this 😅

Yes, I also think that this is stretching this feature too much. When a start and the end date are just dates, the week or day view becomes just a task lists, so I'd be in favor of removing the week/day view if that happens.

Alternatively, there are some other views in the upstream control like task only view https://nhn.github.io/tui.calendar/latest/tutorial-05-weekly-view-no-event-view

Or 2 weeks view on their main page.

berhalak (changes requested):
Please consider using alternative views like the one below, that are design for presenting events that are dates only.

https://nhn.github.io/tui.calendar/latest/tutorial-05-weekly-view-no-event-view
And other examples like 2 weeks view on their team site.

fflorent:
Hi @lmzr,

Do you need help for this PR?

lmzr:
Thanks for the feedback — you were right that the `MutationObserver` was too much. I've reworked this PR to drop it entirely and lean on the library instead.

When both the start and end columns are pure `Date`, the week/day views now switch to TUI's native all-day-only layout via `week.eventView: ['allday']`, so the empty time grid is gone with no CSS hacks. Events are already flagged `isAllday` for pure-date columns, so they land in the all-day panel; the month view is untouched.

One wrinkle: in that mode TUI still reserves layout height for the hidden time grid (and for the day-names row of the other perspective), leaving the all-day panel one row tall and collapsing events behind a "+N more". Instead of clicking exceed buttons in the DOM, I give that space back through TUI's own `weekViewLayout` store (zeroing those rows on `afterRenderEvent`) — the same store this file already uses (`dnd`, `popup`). The all-day panel is the layout's resizable panel, so it absorbs the height and shows every event across view switches, navigation and linked-cursor updates.

Known limitation: if a single day has more events than the panel's full height can fit, TUI falls back to its native "+N more" (it doesn't scroll or render the overflow). That felt like the right place to stop rather than re-growing the layout and managing container scroll — happy to revisit if you'd prefer.

On a broader note: most of the friction here comes from toastui-calendar itself — the all-day overflow, the non-functional "+N" expand, the layout reserving space for hidden rows. The library has had no release since `2.1.3` (Aug 2022) and no meaningful commits since early 2023, so issues like these can only ever be worked around widget-side, never fixed upstream. It might be worth Grist considering a move to an actively maintained calendar library (e.g. FullCalendar) at some point — clearly out of scope for this PR, just flagging it given how much of this widget's complexity traces back to working around that dependency.

Squashed to a single commit, rebased on current master.

PR #206, opened 2026-04-07
Updating dev scripts

berhalak:
- Using `concurrently` npm package to run `grist:serve` and `grist:dev`. This kills both server when stopped without leaving the node still running
- Using explicit versions for dependencies
- Fixing jupiternotebook test


paulfitz (approved):
Thanks @berhalak

Issue #208, opened 2026-04-15
[HTML Viewer] Page doesn't refresh after <a> when editing the mapped cell


0lteron:
## Description:

When the HTML viewer widget displays content containing an `<a href="">` tag, the rendered page in the viewer does not refresh when editing the mapped cell.

## Steps to reproduce:

- Add the HTML Viewer custom widget to a page and map it to a column.
- Put HTML containing an `<a>` tag in the mapped cell, e.g. `<a href="https://www.iamnotarobot.com/">click me</a>`.
- Click on the link of the `<a>` tag in the gadget tab, the viewer should open the link in the iframe.
- Edit the mapped cell to whatever you want (leave it empty  or change the html code).
- The widget still shows the old link, no refresh occurs.

**Expected behavior**: The page gets rendered immediately after the cell edit is confirmed.

Issue #197, opened 2026-02-27
html viewer: internal links should open in the parent window


Ornanovitch:
1. Use the html viewer and display a cell with some internal `<a href...>` links (for eg. with `SELF_HYPERLINK`)
2. Click on the link, it results in a nested grist

<img width="1829" height="960" alt="Image" src="https://github.com/user-attachments/assets/cf7ae689-d2d4-4328-92fc-c4e63258a950" />

Adding `target='_parent'` to the `<a>` does nothing because it is blocked by dompurify

0lteron:
@Ornanovitch  a fix for this would be change the target for each link tag  to _parent after the DOMpurify sanitization, but it should distinguish between internal and external links to keep normal behaviour
i can do it

PR #209, opened 2026-04-18
fix(renderhtml): fix link navigation in HTML viewer widget

0lteron:
## Problem

- Clicking an `<a>` link in the HTML viewer caused the widget iframe to navigate away, destroying the `onRecord` listener.  This meant cell edits no longer refreshed the viewer (#208). 
- Internal Grist links (e.g. from `SELF_HYPERLINK`) would also open a nested Grist instance inside the viewer instead of navigating to the target page (#197).

## Fix

- Intercept all link clicks. External links load in a hidden inner `iframe` that covers the widget area, keeping the outer widget page (and its `onRecord` listener) alive.
- Internal Grist links redirect to the main Grist window directly.

## Changes

- Added a hidden `<iframe id="link-frame">` that covers the widget area
- Added `getGristOrigin` / `isGristLink` to detect internal Grist links using `ancestorOrigins`.
- `hideLinkFrame` is called in `onRecord` and `onNewRecord` so the viewer snaps back to the new rendered HTML on cell edit

### DIFF
--- a/renderhtml/index.html
+++ b/renderhtml/index.html
@@ -19,21 +19,80 @@
           padding: 20px;
           text-align: center;
       }
+      #link-frame {
+          display: none;
+          position: fixed;
+          top: 0;
+          left: 0;
+          width: 100%;
+          height: 100%;
+          border: none;
+      }
     </style>
   </head>
   <body>
     <div id="error"></div>
     <div id="rendered"></div>
+    <iframe id="link-frame" sandbox="allow-scripts allow-same-origin allow-forms allow-popups"></iframe>
     <script>
       var params = new URLSearchParams(window.location.search)
       var ADD_TAGS = params.get('tags')?.split(",");
       var ADD_ATTR = params.get('attr')?.split(",");
+
+      var linkFrame = document.getElementById('link-frame');
+
+      // ancestorOrigins and document.referrer are both readable cross-origin
+      // so they work whether the widget and Grist share an origin or not.
+      function getGristOrigin() {
+        try {
+          if (window.location.ancestorOrigins?.length > 0) {
+            return new URL(window.location.ancestorOrigins[0]).origin;
+          }
+          if (document.referrer) {
+            return new URL(document.referrer).origin;
+          }
+        } catch (_) {}
+        return null;
+      }
+
+      function isGristLink(href) {
+        try {
+          const gristOrigin = getGristOrigin();
+          return gristOrigin !== null && new URL(href).origin === gristOrigin;
+        } catch (_) {
+          return false;
+        }
+      }
+
+      function showLinkFrame(url) {
+        linkFrame.src = url;
+        linkFrame.style.display = 'block';
+      }
+
+      function hideLinkFrame() {
+        linkFrame.style.display = 'none';
+        linkFrame.src = 'about:blank';
+      }
+
       function render(elemId, content) {
         var el = document.getElementById(elemId);
         if (!content) {
           el.style.display = 'none';
         } else {
           el.innerHTML = DOMPurify.sanitize(content, {ADD_TAGS, ADD_ATTR, FORCE_BODY: true});
+          // Intercept link clicks to force this behavior:
+          // - internal grist links redirect you in the main window
+          // - external links load in the inner iframe (keeps the onRecord listener alive).
+          el.querySelectorAll('a[href]').forEach(a => {
+            a.addEventListener('click', function(e) {
+              e.preventDefault();
+              if (isGristLink(a.href)) {
+                window.top.location.href = a.href;
+              } else {
+                showLinkFrame(a.href);
+              }
+            });
+          });
           // If we are allowing scripts, let them execute, which doesn't
           // normally happen when adding script elements using innerHTML.
           if (ADD_TAGS?.includes("script")) {
@@ -60,6 +119,7 @@
         return columns.length === 1 ? record[columns[0]] : undefined;
       }
       grist.onNewRecord(() => {
+        hideLinkFrame();
         render("error", null);
         render("rendered", "");
         lastData = "";
@@ -77,6 +137,7 @@
         } else {
           render("error", null);
           if (lastId !== record.id || lastData !== data) {
+            hideLinkFrame();
             render("rendered", String(data || ''));
           }
           lastId = record.id;
--- a/renderhtml/index.html
+++ b/renderhtml/index.html
@@ -41,15 +41,12 @@
 
       var linkFrame = document.getElementById('link-frame');
 
-      // ancestorOrigins and document.referrer are both readable cross-origin
-      // so they work whether the widget and Grist share an origin or not.
+      // ancestorOrigins is readable cross-origin, so it works whether the widget
+      // and Grist share an origin or not.
       function getGristOrigin() {
         try {
           if (window.location.ancestorOrigins?.length > 0) {
             return new URL(window.location.ancestorOrigins[0]).origin;
-          }
-          if (document.referrer) {
-            return new URL(document.referrer).origin;
           }
         } catch (_) {}
         return null;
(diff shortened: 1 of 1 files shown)


renderhtml/index.html
```diff
@@ -19,21 +19,77 @@
...
       function render(elemId, content) {
         var el = document.getElementById(elemId);
         if (!content) {
           el.style.display = 'none';
         } else {
           el.innerHTML = DOMPurify.sanitize(content, {ADD_TAGS, ADD_ATTR, FORCE_BODY: true});
+          // Intercept link clicks to force this behavior:
+          // - internal grist links redirect you in the main window
+          // - external links load in the inner iframe (keeps the onRecord listener alive).
+          el.querySelectorAll('a[href]').forEach(a => {
```
berhalak: It would be better to attach the click handler on the element itself and just check what was clicked.

renderhtml/index.html
```diff
@@ -19,21 +19,77 @@
...
+          width: 100%;
+          height: 100%;
+          border: none;
+      }
     </style>
   </head>
   <body>
     <div id="error"></div>
     <div id="rendered"></div>
+    <iframe id="link-frame" sandbox="allow-scripts allow-same-origin allow-forms allow-popups"></iframe>
```
berhalak: Wdyt about alternative solution?

- Remove the #rendered div above
- Instead show the iframe
- Render inline the html in the iframe itself using the `srcdoc` attribute 

This way, if the rendered html navigates away (even via javascript function which this solution doesn't support), we still can rerender it when the cell changes.

renderhtml/index.html
```diff
@@ -19,21 +19,77 @@
...
     <script>
       var params = new URLSearchParams(window.location.search)
       var ADD_TAGS = params.get('tags')?.split(",");
       var ADD_ATTR = params.get('attr')?.split(",");
+
+      var linkFrame = document.getElementById('link-frame');
+
+      // ancestorOrigins is readable cross-origin, so it works whether the widget
```
berhalak: But it is fairly new, should we add some fallback (or disable this feature), when this is not available?
https://developer.mozilla.org/en-US/docs/Web/API/Location/ancestorOrigins

berhalak (commented):
Hi @0lteron! Thank you for contribution!
Some comments and questions below, let me know what you think.

renderhtml/index.html
0lteron: Hello! thanks  for checking my PR. 
So basically , i will have to put a click handler that will get attached to the container element, and i will use something like `e.target.closest('a[href]')` to check if the click was on a link?  Does that match what you had in mind?

renderhtml/index.html
0lteron: You are right, actually the first version that i made i had a fallback for firefox (which seems to be the only major browser who adopted it recently), i had used `document.referrer` which i can put back very easily

renderhtml/index.html
0lteron: i didn't think about that, thanks for the advice!
i will try to submit the changess so that you can review them

renderhtml/index.html
0lteron: Regarding your comment about "disabling the feature", do you mean:
- Keep the old behavior: the internal link is opened in the `iframe`.
Or
- Not allow the link to be rendered at all and open them in another tab.

if it's the latter we would have to issue some warning to the user so that he understands what's going on.

0lteron:
@berhalak I've updated the implementation based on your feedback, let me know if anything looks off (since i quickly made the edits)!

Spoffy:
Hey @0lteron! Taken this on from Jarek to free up his time a little.

Had a chat with Paul (our CTO), and I think the thing that would be best is if links in HTML Viewer generally opened up in a new tab in the top-level browser (i.e. escaping the iframe).

You could probably do that by hooking the new(ish) [Navigate API](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate_event) that's in browsers, or adding an `target="_blank"` attribute to anchor links that don't have them (and should).

0lteron:
Hello @Spoffy ! Thanks for taking your time reviewing this
i will check your proposition and try to come up with somethings that aligns with what you said

0lteron:
Hey @Spoffy! i updated the implementation based on your feedback.
- I simplified to just force `target="_blank"` + `rel="noopener noreferrer"` on all links.
- I went with the `target="_blank"` approach which was simpler and easier to implement and also since the navigate event is kinda newish available only on  recent versions of browsers.

renderhtml/index.html
```diff
@@ -19,35 +19,50 @@
...
+          width: 100%;
+          height: 100%;
+          border: none;
+      }
     </style>
   </head>
   <body>
     <div id="error"></div>
-    <div id="rendered"></div>
+    <iframe id="rendered" sandbox="allow-scripts allow-same-origin allow-forms allow-popups"></iframe>
```
Spoffy: question: What's the purpose of nesting the HTML inside another iframe?

renderhtml/index.html
```diff
@@ -19,35 +19,50 @@
...
-            Array.from(el.querySelectorAll("script")).forEach(oldScript => {
-              const newScript = document.createElement("script");
-              Array.from(oldScript.attributes)
-                .forEach(attr => newScript.setAttribute(attr.name, attr.value));
-              newScript.appendChild(document.createTextNode(oldScript.innerHTML));
-              oldScript.parentNode.replaceChild(newScript, oldScript);
-            });
-          }
           el.style.display = 'block';
+          if (elemId === 'rendered') {
```
Spoffy: suggestion: It strikes me as odd to match on `rendered` as the `id`, when this code really only cares that the target element is an iframe. I think it'd be neater and more flexible to check that!

renderhtml/index.html
```diff
@@ -19,35 +19,50 @@
...
+            const sanitized = DOMPurify.sanitize(content, {ADD_TAGS, ADD_ATTR, FORCE_BODY: true});
```
Spoffy: suggestion: This can be lifted above the `if` block, so that it's not duplicated here and on line 61, preventing situations where one of these is modified and the other isn't.

Spoffy (changes requested):

Spoffy:
Thanks for making those changes @0lteron, I've left a few comments with some questions and suggestions :)

renderhtml/index.html
0lteron: Hello ! thanks for your review, i hope i understood your question:
 
The iframe was originally added to prevent link clicks from navigating the widget's own page away (which would destroy the `onRecord` listener, that was originally in my first PR). However, since all links now get `target="_blank"`, that's indeed no longer needed.

I'll simplify it back to a `<div>` as it was before.

renderhtml/index.html
0lteron: very much so, maybe using `instance of HTMLIFrameElement` would be better (now it's useless since i will switch to `<div>`, but thanks for pointing this out! )

renderhtml/index.html
0lteron: I will do that immediately thanks!

0lteron:
Hi @Spoffy, thank you for your review, i have submitted a new version with the suggestions you have provided me !

renderhtml/index.html
```diff
@@ -28,26 +31,26 @@
...
-          // normally happen when adding script elements using innerHTML.
-          if (ADD_TAGS?.includes("script")) {
-            Array.from(el.querySelectorAll("script")).forEach(oldScript => {
-              const newScript = document.createElement("script");
-              Array.from(oldScript.attributes)
-                .forEach(attr => newScript.setAttribute(attr.name, attr.value));
-              newScript.appendChild(document.createTextNode(oldScript.innerHTML));
-              oldScript.parentNode.replaceChild(newScript, oldScript);
-            });
-          }
```
Spoffy, line 47: Looks like we've lost this feature in your change, can we restore it to avoid anyone currently using this feature from being disappointed? 🙂

Spoffy (changes requested):

renderhtml/index.html
0lteron: Oh yes, my bad, I didn't notice it. It should be re-added now.

Spoffy: Did you accidentally revert all the changes @0lteron? :grin: 

The diff viewer for the PR now shows no changes :)

PR #210, opened 2026-04-23
Adds a NPM publish CI action gated by widget version bumps

Spoffy:
There's currently no workflow for publishing a new NPM package with bundled Grist widgets.

This sets up a new Github action that:

1. Checks if any widgets included in the NPM package have had their versions bumped, and does no further work if not.
2. Updates the main package.json's version number based on the changes to widget version numbers.
  - Bumps the major version if a widget has had a major version bump, minor if there's only minor bumps, etc.
3. Publishes the bundle to NPM

This is a manually triggered workflow, with the future possibility of being run on any push to master.


.github/workflows/publish-npm-package.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Install dependencies
+        run: yarn install --frozen-lockfile
+
+      - name: Get desired version
+        id: get_version
+        run: echo "version=$(node buildtools/get_version.js --update-cache)" >> "$GITHUB_OUTPUT"
+
+      - name: Get current version
+        id: current_version
+        run: echo "version=$(node -p "require('./package.json').version")" >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
Spoffy: Fixed

buildtools/get_version.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+function determineBumpType(oldVer, newVer) {
+  const oldSv = parseSemver(oldVer);
+  const newSv = parseSemver(newVer);
+  if (!oldSv || !newSv) { return BUMP_MAJOR; }
+  if (newSv.major !== oldSv.major) { return BUMP_MAJOR; }
+  if (newSv.minor !== oldSv.minor) { return BUMP_MINOR; }
+  if (newSv.patch !== oldSv.patch) { return BUMP_PATCH; }
+  return BUMP_NONE;
```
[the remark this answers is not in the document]
Spoffy: Fixed - it now errors on an unparseable version

.github/workflows/publish-npm-package.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: echo "version=$(node buildtools/get_version.js --update-cache)" >> "$GITHUB_OUTPUT"
+
+      - name: Get current version
+        id: current_version
+        run: echo "version=$(node -p 'require("./package.json").version')" >> "$GITHUB_OUTPUT"
+
+      - name: Bump version if changed
+        if: steps.get_version.outputs.version != steps.current_version.outputs.version
+        run: |
+          npm version "$VERSION" --no-git-tag-version
```
paulfitz, line 65: Why no git tag?

.github/workflows/publish-npm-package.yml
```diff
@@ -0,0 +1,54 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "github-actions[bot]@users.noreply.github.com"
+          git add package.json buildtools/version-cache.json
+          git commit -m "Bump version to $VERSION"
+          git push
```
paulfitz: Have you checked this sequence works ok in a fork?

.github/workflows/publish-npm-package.yml
```diff
@@ -0,0 +1,54 @@
...
+        env:
+          VERSION: ${{ steps.get_version.outputs.version }}
+
+      - name: Publish to npm
```
paulfitz, line 85: I note that the package.json change up there will get pushed even if the npm publishing fails.

.github/workflows/publish-npm-package.yml
```diff
@@ -0,0 +1,54 @@
...
+        if: steps.get_version.outputs.version != steps.current_version.outputs.version
+        run: npm publish --access public $DRY_RUN
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+          DRY_RUN: ${{ github.ref_name != 'master' && '--dry-run' || '' }}
```
paulfitz: Could this be a workflow flag rather than hardcoded to a branch?

buildtools/get_version.js
```diff
@@ -0,0 +1,135 @@
...
+
+  // Read current root version.
+  const rootPkg = JSON.parse(fs.readFileSync(path.join(rootDir, 'package.json'), 'utf-8'));
+  const currentVersion = rootPkg.version;
+
+  // Compute new version.
+  const newVersion = worstBump ? applyBump(currentVersion, worstBump) : currentVersion;
+
+  // Output the desired version.
+  process.stdout.write(newVersion);
```
paulfitz: How about a newline as well?

.github/workflows/publish-npm-package.yml
paulfitz: Consider reordering, or doing more to add confidence the publish will work earlier (at least build).

paulfitz (commented):
I think the basic idea works, some comments.

.github/workflows/publish-npm-package.yml
Spoffy: Good catch - holdover from an earlier design I had for this that used two actions.

.github/workflows/publish-npm-package.yml
Spoffy: I haven't, I'll give that a go!

.github/workflows/publish-npm-package.yml
Spoffy: Updated to add a git tag pointing to the version bump commit, that can be skipped if just testing the flow :slightly_smiling_face:

.github/workflows/publish-npm-package.yml
Spoffy: Works fine on a fork, and on other branches :slightly_smiling_face:

.github/workflows/publish-npm-package.yml
Spoffy: I've added a prepack step to at least check the bundle can be built at the current commit. I'm open to running the full test suite too if you feel that'd be worthwhile!

.github/workflows/publish-npm-package.yml
Spoffy: Done - it's now a workflow option :slightly_smiling_face:

buildtools/get_version.js
Spoffy: I'll peek at this tomorrow - not sure if I'd need to trim the output before using it with a newline.

buildtools/bundle.js
```diff
@@ -97,20 +98,23 @@ class Bundler {
   // subdirectories can be crawled correctly) but might be a little
   // unexpected.
   async wait() {
-    let ct = 0;
-    while (true) {
+    let foundServer = false;
+    while (this.localServer.pid && !this.localServer.exitCode) {
```
paulfitz: Could be worth checking also `subprocess.signalCode` although it shouldn't matter when run in CI

buildtools/get_version.js
```diff
@@ -0,0 +1,135 @@
...
+
+const BUMP_NONE  = 0;
+const BUMP_PATCH = 1;
+const BUMP_MINOR = 2;
+const BUMP_MAJOR = 3;
+// Maps severity level back to a bump type string for applyBump().
+const BUMP_TYPES = [null, 'patch', 'minor', 'major'];
+
+function parseSemver(str) {
+  const m = /^(\d+)\.(\d+)\.(\d+)$/.exec(String(str || ''));
```
paulfitz: What do you think about using `semver`? In case anyone does anything weird? We do control the repo though so I see that it's not essential.

.github/workflows/publish-npm-package.yml
```diff
@@ -0,0 +1,92 @@
...
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: 'Perform a dry run (skip actual publish)'
+        type: boolean
+        default: true
+      skip_tag:
+        description: 'Skips adding a git tag for the version'
+        type: boolean
+        default: true
```
paulfitz: Defaulting to skipping a git tag seems unexpected?

paulfitz (commented):
Looks reasonable! A few comments.

.github/workflows/publish-npm-package.yml
Spoffy: My thinking was to not make any changes by default which escape from the branch (tags, npm publish). Running this branch without changing any flags will only make a new commit.

I can change that though if you think it's going to cause confusion :)

buildtools/bundle.js
Spoffy: Good idea. I thought .signalCode tracked every signal, but I just checked and it's apparently only terminating signals, so seems safe to do :slightly_smiling_face:

buildtools/get_version.js
Spoffy: Done, switched it all to using the semver lib, seems like an overall improvement in reliability :)

.github/workflows/publish-npm-package.yml
paulfitz: Isn't that what dry-run should do? And if not a dry-run, shouldn't the run be fully effective by default?

Ah, I see, you want dry-run to be strictly for the npm publish step?

Ok I don't think any of this is that important, all I want is a way to run this branch effectively (make a release, including a tag and a package). The flags you chose feel a little easy to cut yourself on, but then the person trying to publish can just do it again and read more until you get it right.

paulfitz (approved):

PR #213, opened 2026-04-29
Wait for the test Grist server to start up cleanly

paulfitz:
Two small tweaks so our tests stop tripping over a fresh Grist server:

- We now set GRIST_IN_SERVICE=1 when we start the container, so Grist knows it's good to go and doesn't ask us to visit a setup page.
- We now wait on /status?ready=1, which only says "OK" once Grist is truly ready to take requests. The plain /status was answering "I'm alive!" a few seconds before it could actually do anything, and our uploads were sneaking in too early and getting turned away.


berhalak (approved):
Thank you @paulfitz !

PR #218, opened 2026-05-26
Moves all dependencies to devDependencies

Spoffy:
http-server and jsonc are only used in development scripts.

Move them to devDependencies so grist-widget has no dependencies when bundled.


paulfitz (approved):
Thanks Callum. I see tests are failing. I've opened a PR to update.

PR #221, opened 2026-06-18
Show widget labels as text instead of HTML

paulfitz:
A few widgets were dropping cell values straight onto the page as HTML. Usually that's harmless, but it means a stray "<" or a snippet of markup in a label could show up in surprising ways instead of as the text someone actually typed. This nudges them to show text as text.

- Map: popup labels now go through DOMPurify before they're shown (Leaflet treats popup content as HTML).
- Flashcards: the question and answer are set with textContent, and a list-style answer becomes a simple bullet list.
- Timeline: the detail drawer builds its rows as DOM nodes with textContent rather than gluing together an HTML string.

Adds a couple of browser tests (Map and Flashcards) that feed in a label containing some HTML and check it shows up as plain text, and rebuilds the Timeline bundle.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,8 @@ on:
 jobs:
   build_and_test:
     runs-on: ubuntu-latest
+    # Backstop against a hung run.
+    timeout-minutes: 20
     strategy:
       matrix:
         python-version: [3.9]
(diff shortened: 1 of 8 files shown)


fflorent (approved):
LGTM, thank you!

PR #224, opened 2026-06-19
Render images in HTML markdown widget

fflorent:
The user may use `__ATTACHMENT[id]__` to reference the URL of the attachment, that can be used to render images.

Though I don’t think it is very discoverable, so pretty much sure it won’t be actually used…

Here is a preview of the widget in action: https://docs.getgrist.com/e3H1GJRYrfab/Mardown-with-images?utm_id=share-doc


fflorent:
TODO for Florent: improve description, with an example, and how the markdown is transformed. Also give a preview if possible.

fflorent:
Converting to draft.

I wanted this feature at a specific moment and made this development. But this is far from being complete. Also I have taken a second look at the widget and at the EasyMDE library, there is an editor which offers a way to [customize its toolbar](https://github.com/Ionaru/easy-markdown-editor#toolbar-customization). I think this could solve the issue of the discoverability. 

I think of starting again from the need

PR #227, opened 2026-06-19
Render inspect error and exoplanet greeting as text

paulfitz:
Both write plain text through innerHTML for no reason: inspect/api.js shows an error message, and exoplanet shows a greeting built from a record's name. Use textContent. Same cleanup as #225, for the two spots it didn't cover.


fflorent (approved):
Thanks!

PR #230, opened 2026-06-23
Switch to OSM tiles by default (?)

fflorent:
A user (ecouillard) asked us why we don't provide tiles from OSM for the Map widget by default:
https://forum.grist.libre.sh/t/widget-carte-possibilite-dafficher-le-fond-de-carte-plan-ign/3873/2?u=florent.fayolle

While experimenting the use of OSM, I felt like the latter is slightly smoother to load tiles (for example while playing with the zoom) than the current service (arcgisonline).
Both don't display the same elements, but I feel like it's okay-ish? (and I like the look-and-feel as well as the details from OSM).

If this PR does not retain your attention, I would be very curious to understand the motivations for choosing arcgisonline :)

## Before (with arcgisonline)

Map using arcgisonline.com

## After (using OSM)

Map widget using OSM

### DIFF
--- a/map/page.js
+++ b/map/page.js
@@ -8,8 +8,8 @@ let selectedTableId = null;
 let selectedRowId = null;
 let selectedRecords = null;
 let mode = 'multi';
-let mapSource = 'https://server.arcgisonline.com/ArcGIS/rest/services/World_Street_Map/MapServer/tile/{z}/{y}/{x}';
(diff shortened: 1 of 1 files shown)


georgegevoian:
The last time the source changed was in this PR: https://github.com/gristlabs/grist-widget/pull/53. There were problems with the previous source at higher zoom levels.

I think I also prefer OSM slightly to the current default. I have no issue with changing it, but it is a subjective change and there may be users who prefer the old one. It's a little tricky because changes to the widget land immediately, and there isn't a natural place to put a notice and instructions on reverting to the previous source. I don't recall the last change to the source being particularly controversial, so it might not be worth worrying about.

Let's leave this open a little longer for others to chime in. If there's no major pushback, I think it's ok to change.

paulfitz (approved):
Let's give it a try. If there's pushback, we may have to think of some cleverer way to manage this kind of change.

PR #231, opened 2026-07-02
fix(calendar): correct mistyped toastui template keys so translations apply

lmzr:
Two template callbacks in the calendar widget were registered under **misspelled keys**, so toastui-calendar never invoked them and rendered its hardcoded English labels:

- `poupSave` → `popupSave` (event popup "Save" button)
- `allDayTitle` → `alldayTitle` (week/day all-day row label)

The i18next keys (`t('Save')`, `t('All Day')`) already existed in every populated translation file; only the template names were wrong. Bug present since #65.

**Follow-up (2nd commit):** once `alldayTitle` fires, it returns a *string*, which toastui wraps in a bare `.template-alldayTitle` element (it only merges the class onto its own default *vnode*). That dropped the built-in `.left-content` cell styling, leaving the label oversized and top-aligned. A CSS rule re-applies `display:table-cell; font-size:11px; text-align:right; vertical-align:middle` to the wrapper, restoring the default look — now translated.

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -187,7 +187,7 @@ class CalendarHandler {
         popupDelete(){
           return t('Delete')
         },
-        poupSave(){
+        popupSave(){
           return t('Save')
         },
         popupEdit(){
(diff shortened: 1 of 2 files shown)


fflorent (approved):
LGTM

PR #232, opened 2026-07-07
fix(calendar): correct first day of week on browsers without Intl getWeekInfo

lmzr:
## Problem

`getFirstDayOfWeek()` relies solely on `Intl.Locale#getWeekInfo()` /
`#weekInfo` to decide the first day of the week. Firefox implements neither
(any version), so the API returns `undefined` and the calendar falls back to
**Sunday** — wrong for France and most of the world. Reported on the Grist
community forum (Firefox ESR 140): the week started on Sunday for `fr-FR`.

## Fix

Add a data-driven `FIRST_DAY_BY_REGION` table derived directly from CLDR
supplemental `weekData/firstDay` (56 Sunday-first, 14 Saturday-first, 1
Friday-first region; everything else defaults to Monday). When
`getWeekInfo`/`weekInfo` is unavailable, resolve the region via
`Intl.Locale#maximize()` (supported in Firefox) and look it up in the table.
The ultimate fallback changes from Sunday to **Monday**, the CLDR world
default (`firstDay "001"`).

The primary `getWeekInfo` path (Chrome/Safari) is unchanged, so behaviour is
identical where the API exists. Also fixes a misleading comment that claimed
Firefox 126+ supports `weekInfo`.

## Testing

- Node test across 16 locales (fr-FR→Mon, en-US→Sun, ar-EG→Sat, dv-MV→Fri,
  pt-PT→Sun per CLDR, unknown→Mon, …): all pass.
- In-browser: Vivaldi (primary path, no regression) and Firefox (fallback
  path) — `fr-FR` now starts the week on Monday in Firefox.

### DIFF
--- a/calendar/page.js
+++ b/calendar/page.js
@@ -53,19 +53,49 @@ function getMonthName() {
   return calendarHandler.calendar.getDate().toDate().toLocaleString(getLanguage(), {month: 'long', year: 'numeric'})
 }
 
+// First day of week per CLDR region, for browsers that do NOT implement
+// Intl.Locale#getWeekInfo / #weekInfo (notably every version of Firefox).
+// Values are TUI day codes (0=Sun, 5=Fri, 6=Sat). Any region absent from this
+// table starts the week on Monday — the world default (CLDR firstDay "001").
+// Derived directly from CLDR supplemental weekData/firstDay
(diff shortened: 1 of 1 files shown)


fflorent (approved):
LGTM, needs a GristLabs approval, I track it on our board so it is not forgotten.

PR #235, opened 2026-08-06
Pushing column mapping definition to widget's manifest

berhalak:
Column mappings live in the widget code, inside `grist.ready()`. Grist only
sees them after the widget loads. An assistant that adds a widget to a page
cannot see them at all, so it cannot map the columns.

This PR adds the mappings to `package.json`, so they end up in
`manifest.json`. This is the new place to define columns for a widget.

The old mappings in `grist.ready()` stay, so nothing breaks. New widgets can
use the manifest only.

Not related: `rendervideo` used `label` instead of `title`, now fixed.


inspect/package.json
```diff
@@ -70,7 +70,11 @@
           "url": "https://github.com/gristlabs"
         }
       ],
-      "isGristLabsMaintained": true
+      "isGristLabsMaintained": true,
+      "columns": [
+        "Link",
+        "Title"
```
paulfitz: Is this bad copy/paste?

paulfitz (commented):
Thanks @berhalak, looks pretty good.

In passing, could you fix the existing issues of a double grist.ready in flashcards:
```
  grist.ready({
    columns: [
      { name: "Question", type: 'Text', title: "Question Column"},
      { name: "Answer", type: 'Text', title: "Answer Column"},
    ],
    requiredAccess: 'read table'
  });
  grist.ready();
```
and in maps/page.js there's a different duplication:
```
    { name: "Address", type: 'Text', optional, optional},
```

inspect/package.json
berhalak: Yes, my mistake, I wanted to remove this modification from this widget.

paulfitz (approved):
thanks @berhalak !

